PR #4, opened 2016-12-09
SAR-111: Added session timeout view, controller and action as well as relevant tests.

JSJohal:


app/views/timeout/timeout.scala.html
```diff
@@ -0,0 +1,5 @@
+@()(implicit request: Request[_], messages: Messages))
+
+@views.html.main_template(title = "Session timeout", bodyClasses = None) {
```
MJCallahanPage: `title = "Session timeout"`
Instead of doing this can you use the Messages file to setup a Timeout.title value and use that in the view. 

E.g.: `title = "@messages.Timeout.title"` - see other frontend repos for more examples.

app/views/timeout/timeout.scala.html
```diff
@@ -0,0 +1,5 @@
...
+    <h1>Your session has timed out.</h1>
```
MJCallahanPage: Same here for the heading. Create something in the Messages file like Timeout.heading. then make the code:

`<H1>@messages.Timeout.heading</H1>`

app/controllers/HelloWorldController.scala
```diff
@@ -26,6 +26,6 @@ object HelloWorldController extends HelloWorldController
 
 trait HelloWorldController extends FrontendController {
   val helloWorld = Action.async { implicit request =>
-		Future.successful(Ok(views.html.helloworld.hello_world()))
+		Future.successful(Ok(views.html.hello_world()))
```
MJCallahanPage: Can you undo this change so that the hello_world view is still in the `helloworld` package.

Just keeps it separate - so that we can remember to remove it all later.

app/views/timeout/timeout.scala.html
```diff
@@ -0,0 +1,5 @@
...
+}
```
MJCallahanPage: Should the timeout view have a restart type button? So that they can make a positive action to restart their journey? 

What have other services done?

jsjohal:
@MJCallahanPage Made changes as recommended

PR #14, opened 2017-01-06
Sar 173

suyangyu:


app/forms/DateForm.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import models.DateModel
+import play.api.data.Form
+import play.api.data.Forms._
+
+object DateForm {
+
+  val dateDay = "dateDay"
+  val dateMonth = "dateMonth"
+  val dateYear = "dateYear"
```
MJCallahanPage: is there any need to have these vals declared? They are only used once in the mapping?

Can we just use the string direct?

app/forms/EmailForm.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import models.EmailModel
+import play.api.data.Form
+import play.api.data.Forms._
+
+object EmailForm {
+
+  val emailForm = Form(
+    mapping(
+      "emailAddress" -> text
```
MJCallahanPage: This is how I thought we'd have it. Unless there's a valid reason for the 'val' declarations?

app/forms/IncomeTypeForm.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package forms
+
+import models.IncomeTypeModel
+import play.api.data.Form
+import play.api.data.Forms._
+
+object IncomeTypeForm {
+
+  val incomeType = "incomeType"
```
MJCallahanPage: This val is declared and then never used; can we remove it.

app/forms/TermForm.scala
```diff
@@ -0,0 +1,32 @@
...
+import models.TermModel
+import play.api.data.Form
+import play.api.data.Forms._
+
+
+object TermForm {
+
+  val termForm = Form(
+    mapping(
+      "termName" -> boolean
```
MJCallahanPage: Can this be something like "hasAcceptedTerms" so that it is clear what this boolean value represents.

app/forms/DateForm.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  val dateForm = Form(
+    mapping(
+      dateDay -> text,
+      dateMonth -> text,
+      dateYear -> text
+    )(DateModel.apply)(DateModel.unapply)
+  )
+}
```
MJCallahanPage: How is this anticipated to work now that we have two dates on the one page?

app/models/TermModel.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import play.api.libs.json.Json
+
+case class TermModel(agreeToTerms: Boolean)
```
MJCallahanPage: This parameter name must match the name in the form. So that it can bind to the form successfully without having to write our own formatters. I.e. the model.apply model.unapply will work.

So I suggest making this `hasAcceptedTerms` and making the form match. Or, change the form to `agreeToTerms` so that it matches.

app/forms/DateForm.scala
suyangyu: It's a good idea to have them as constants because we can then get IDE support when we need them. A couple of places where we already use them would be to set up test data.

In other examples of where they might be required is when we want a specific field from a form.

app/forms/DateForm.scala
suyangyu: It currently doesn't
we'll need to change it to a submapping

app/forms/TermForm.scala
suyangyu: ok

app/models/TermModel.scala
suyangyu: will make the amendments.

However for the record the parameter name doesn't need to match the form.
In play the name of the form must match the name used in the mapping.
how that get put into the case class is dependent on how they're ordered in the form.

e.g.
case class(name1:String,name2:String)
mapping ("name2" -> text, "name1" -> text)...
will in fact populate the case class incorrectly

app/forms/DateForm.scala
MJCallahanPage: agreed, we can refactor when implementing the view.

app/forms/DateForm.scala
MJCallahanPage: Let's be consistent across all forms then :)

app/forms/TermForm.scala
```diff
@@ -0,0 +1,33 @@
...
+import play.api.data.Form
+import play.api.data.Forms._
+
+
+object TermForm {
+  val termName = "hasAcceptedTerms"
+
+  val termForm = Form(
+    mapping(
+      termName -> boolean
```
MJCallahanPage: can we change val name to match as per the other forms :)

app/forms/IncomeTypeForm.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  val incomeTypeForm = Form(
+    mapping(
+      "incomeType" -> text
```
MJCallahanPage: Can we use the val here rather than the string.

PR #15, opened 2017-01-09
Error message library for defining and converting to String

suyangyu:


test/forms/validation/ErrorMessageHelperSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  val testInvalid = ErrorMessageFactory.error("errMsg", "arg1", "arg2")
+
+  val checkLength: String => ValidationResult = (a: String) =>
+    a.length > 10 match {
+      case true => Valid
+      case false => testInvalid
+    }
+
+  def testConstraint[A](f: A => ValidationResult) = Constraint[A]("")(f)
```
MJCallahanPage: Can you refactor this into some sort of constraint util/helper/sugar so that it is part of the application and an be used when we implement the validation on the forms. 

Then you can use that util/helper/sugar in this test without needing to definine it.

suyangyu: yes, will do

PR #44, opened 2017-02-01
Sar 299

suyangyu:


app/auth/AuthorisedForIncomeTaxSA.scala
```diff
@@ -47,20 +54,32 @@ trait AuthorisedForIncomeTaxSA extends Actions {
...
+          implicit request =>
+            checkEnrolment {
+              case NotEnrolled => action(IncomeTaxSAUser(authContext))(request)
+              case Enrolled => Future.successful(Redirect(alreadyEnrolledUrl))
+            }
       }
-    }
   }
 
+  def checkEnrolment(f: Enrolled => Future[Result])(implicit hc: HeaderCarrier): Future[Result] =
```
MJCallahanPage: Can I suggest we create an 'Enrolment' service in our microservice. I.e. under `app\services` package.

That way, we can include all the enrolment functionality within that going forward :smile: 

For example:
- addKnownFacts
- createEnrolment

Which will come later :)

MJCallahanPage:
Merging as is, with a tech task to refactor later.

PR #55, opened 2017-02-21
SAR-352

suyangyu:
added the test only de enrol function and refactored appConfig + enrollment functions

### DIFF
--- a/app/connectors/EnrolmentConnector.scala
+++ b/app/connectors/EnrolmentConnector.scala
@@ -18,29 +18,24 @@ package connectors
 
 import javax.inject.{Inject, Singleton}
 
+import config.AppConfig
 import connectors.models.Enrolment
-import play.api.Application
 import play.api.http.Status._
-import uk.gov.hmrc.play.config.ServicesConfig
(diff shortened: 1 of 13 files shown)


app/connectors/EnrolmentConnector.scala
```diff
@@ -50,10 +45,8 @@ class EnrolmentConnector @Inject()(override val app: Application,
 
 object EnrolmentConnector {
 
-  val authorityUri = "auth/authority"
-
   //TODO update once constant is confirmed
```
MJCallahanPage: Can you remove the TODO comment as this is now done.

app/connectors/EnrolmentConnector.scala
```diff
@@ -50,10 +45,8 @@ class EnrolmentConnector @Inject()(override val app: Application,
...
-  val enrolmentOrgKey = "SABR"
+  val enrolmentKey = "HMRC-MTD"
   //TODO update once constant is confirmed
```
MJCallahanPage: Same here

suyangyu: done for both

PR #71, opened 2017-03-06
SAR-400: Added back URL for terms view and relevant tests

JSJohal:


app/controllers/TermsController.scala
```diff
@@ -19,42 +19,59 @@ package controllers
...
+  def backUrl(implicit request: Request[_]): Future[String] =
+    keystoreService.fetchIncomeSource() flatMap {
+      case Some(source) => source.source match {
+        case IncomeSourceForm.option_business =>
+          controllers.business.routes.BusinessIncomeTypeController.showBusinessIncomeType().url
+        case IncomeSourceForm.option_both =>
+          controllers.business.routes.BusinessIncomeTypeController.showBusinessIncomeType().url
+        case IncomeSourceForm.option_property =>
+          controllers.routes.IncomeSourceController.showIncomeSource().url
+        case _ => new InternalServerException("Internal Server Error")
```
suyangyu: Can you make this a bit more meaningful, although it shouldn't happen it'll make it easier to debug if it does occur

JSJohal: Change made.

PR #74, opened 2017-03-07
Sar 416

suyangyu:


app/controllers/HomeController.scala
```diff
@@ -19,17 +19,33 @@ package controllers
...
     implicit request =>
-      Redirect(controllers.preferences.routes.PreferencesController.checkPreferences())
+      baseConfig.applicationConfig.enableThrottling match {
+        case true =>
+          throttlingService.checkAccess.flatMap {
+            case Some(CanAccess) =>
+              gotoPreferences
+            case Some(_) =>
+              // TODO show the page
+              Ok("exceeds limit")
```
MJCallahanPage: Do we not have a design for the throttle exceeded page yet? Would be good to get that in now if we can. Speak to Anthony, Rohima or Oli

suyangyu: I have that on a separate subtask :)

MJCallahanPage: Good, at least it's being tracked. Will merge :)

PR #75, opened 2017-03-07
SAR-397: Created view, controllers and models for other-income page, for other-income-error page, added messages and routes. Also updated main income error view to include back URL and updated main income error controller test as spotted a mistake

JSJohal:
…for other-income-error page, added messages and routes. Also updated main income error view to include back URL and updated main income error controller test as spotted a mistake


app/controllers/OtherIncomeErrorController.scala
```diff
@@ -0,0 +1,42 @@
...
+class OtherIncomeErrorController @Inject()(implicit val baseConfig: BaseControllerConfig,
+                                           val messagesApi: MessagesApi) extends BaseController {
+
+  val showOtherIncomeError = Action.async { implicit request =>
+    Future.successful(Ok(views.html.other_income_error(postAction = controllers.routes.OtherIncomeErrorController.submitOtherIncomeError(), backUrl)))
+  }
+
+  val submitOtherIncomeError: Action[AnyContent] = Authorised.async { implicit user =>
+    implicit request =>
+      Future.successful(Redirect(controllers.business.routes.BusinessAccountingPeriodController.showAccountingPeriod()))
```
MJCallahanPage: I think this route might be dependant on what they have selected upfront as I don't think the prototype is correct and Oli agrees with me.

If they select property only they should still see the other income screen, which means that they might go to T&Cs instead of Accounting Period.

app/controllers/OtherIncomeController.scala
```diff
@@ -0,0 +1,70 @@
...
+    implicit request =>
+      OtherIncomeForm.otherIncomeForm.bindFromRequest.fold(
+        formWithErrors => BadRequest(view(otherIncomeForm = formWithErrors, backUrl = backUrl)),
+        choice =>
+          keystoreService.saveOtherIncome(choice).flatMap { _ =>
+            choice.choice match {
+              case OtherIncomeForm.option_yes =>
+                Redirect(controllers.routes.OtherIncomeErrorController.showOtherIncomeError())
+              case OtherIncomeForm.option_no =>
+                Redirect(controllers.business.routes.BusinessAccountingPeriodController.showAccountingPeriod())
```
MJCallahanPage: Same as one of my other comments.

This might need to go to T&Cs if it's the property only route. Need to speak to Rohima, Anthony and Oli tomorrow.

conf/messages
```diff
@@ -197,10 +200,34 @@ not-eligible.signup                                             = I want to sign
...
+other-income-error.para2                                        = You can send digital Income Tax updates for your other income later in the year.
+other-income-error.bullet1                                      = a sole trader with income from one business
+other-income-error.bullet2                                      = someone who lets UK property
+other-income-error.bullet3                                      = a sole trader with income from one business and you let a UK property
+
+## Other sources of income page ##
+income-other.title                                              = Do you have any other sources of income?
+income-other.heading                                            = Do you have any other sources of income?
+income-other.para1                                              = This includes:
+income-other.bullet1                                            = employment that isn’'t your sole trader business
```
MJCallahanPage: This should be double signle quote. This looks like a special character and then a single quote.

conf/messages
```diff
@@ -197,10 +200,34 @@ not-eligible.signup                                             = I want to sign
...
+income-other.bullet2                                            = UK pensions or annuities
+income-other.bullet3                                            = state benefits
+income-other.bullet4                                            = student loans
+income-other.bullet5                                            = employment or investments from outside the UK
+income-other.bullet6                                            = capital gains
+income-other.bullet7                                            = the ‘Rent a Room Scheme’
```
MJCallahanPage: Instead of these special characters for the quotation use double single quote.

MJCallahanPage:
Raising a separate Spike activity to investigate the flow issues.

PR #78, opened 2017-03-08
added test only route for reseting the throttle

suyangyu:


app/controllers/HomeController.scala
```diff
@@ -40,7 +40,7 @@ class HomeController @Inject()(override val baseConfig: BaseControllerConfig,
               gotoPreferences
             case Some(_) =>
               // TODO show the page
```
MJCallahanPage: remove todo

MJCallahanPage:
Also, quick fix needed on conflicts :)

PR #79, opened 2017-03-08
SAR-401: Removed controllers, views, forms and models which were part of proto…

JSJohal:
…type v1 but are no longer required. Also reflected those changes in KeystoreService, routes and updated relevent tests


MJCallahanPage:
@JSJohal - can you fix the conflicts please. Cheers.

app/views/summary_page.scala.html
```diff
@@ -83,7 +83,7 @@ <h1 class="heading-xlarge">@Messages("summary.heading")</h1>
         @summaryModel.accountingPeriod.map(accountingPeriod)
         @summaryModel.businessName.map(businessName)
         @summaryModel.incomeType.map(incomeType)
-        @summaryModel.contactEmail.map(contactEmail)
+@*//        @summaryModel.contactEmail.map(contactEmail)*@
```
MJCallahanPage: Can you delete this; rather than comment it out.

Also, can you delete this section here: https://github.com/hmrc/income-tax-subscription-frontend/blob/faa77373027fd9513585c8c00836d45b91baab93/app/views/summary_page.scala.html#L54-L60

PR #131, opened 2017-04-18
Sar 564

suyangyu:
**This PR cannot be merged into master**

This change requires end-to-end testing in QA with DES and ETMP. This needs to be lined up.


app/auth/AuthorisedForIncomeTaxSA.scala
```diff
@@ -87,7 +87,10 @@ trait AuthorisedForIncomeTaxSA extends Actions with ErrorPageRenderer {
       authedBy.async {
         authContext: AuthContext =>
           implicit request =>
-            action(IncomeTaxSAUser(authContext))(request).flatMap { x => x.withSession(x.session.+(GoHome -> "et")) }
+            enrolmentService.checkEnrolment {
```
MJCallahanPage: Do we not need a check for the `subscriptionService`? Rather than the `enrolmentService`?

app/controllers/AlreadyEnrolledController.scala
```diff
@@ -31,7 +31,7 @@ class AlreadyEnrolledController @Inject()(val baseConfig: BaseControllerConfig,
                                          val messagesApi: MessagesApi
                                         ) extends BaseController {
 
-  val enrolled = Authorised.asyncForEnrolled {  implicit user =>
+  val enrolled = Authorised.asyncCore {  implicit user =>
```
MJCallahanPage: This method should be called `subscribed` not `enrolled`

They are similar, and yet, very different things. :smile:

app/controllers/AlreadyEnrolledController.scala
```diff
@@ -31,7 +31,7 @@ class AlreadyEnrolledController @Inject()(val baseConfig: BaseControllerConfig,
...
     implicit request =>
       Future.successful(Ok(views.html.enrolled.already_enrolled(postAction = controllers.routes.SignOutController.signOut())))
```
MJCallahanPage: Might be worth adding a comment here that we are re-using the already-enrolled view for now but it might be replaced with already-subscribed if the content changes.

MJCallahanPage:
**This PR cannot be merged into master**

This change requires end-to-end testing in QA with DES and ETMP. This needs to be lined up.

PR #135, opened 2017-05-08
added splunk audit for survey

suyangyu:


app/controllers/ExitSurveyController.scala
```diff
@@ -49,13 +53,35 @@ class ExitSurveyController @Inject()(val app: Application,
...
     )
   }
 
-  //TODO replace with calls to the persistent backend once it has been decided
-  def submitSurvey(survey: ExitSurveyModel): Future[Boolean] = {
-    Future.successful(true)
+  def submitSurvey(survey: ExitSurveyModel)(implicit hc: HeaderCarrier): Unit = {
+    val surveyAsMap = surveyFormDataToMap(survey)
+    if (surveyAsMap.nonEmpty)
+      logging.audit(transactionName = "ITSA Survey", detail = surveyAsMap, auditType = Logging.eventTypeSuccess)
```
MJCallahanPage: I assume the transaction name relates to the 'auditType' in splunk. Can we change this so that it is in line with other audit events we do. i.e.: `mtdItsaExitSurveyFeedback`

Can you also create a page on Confluence for our Splunk events and document this similar to how the transaction monitoring is documented.

PR #244, opened 2017-10-03
Sar 976

suyangyu:


app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -15,24 +15,37 @@
...
 sealed trait FeatureSwitch {
   val name: String
+  val displayText: String
 }
 
 object FeatureSwitch {
   val prefix = "feature-switch"
+
+  def apply(str: String): FeatureSwitch =
+    switches.filter(_.name == str).toList match {
```
kabuumu: Could this be:

`switches find (_.name == str) match {`
`  case Some(switch) => switch `
`  case None => throw new IllegalArgumentException("Invalid feature switch: " + str)`
`}`

PR #335, opened 2018-01-10
Update reset profile call to use government-gateway-authentication/re…

tonytw1:
…fresh-profile rather that the deprecated authenticator/refresh-profile.

This is safe because authenticator/refresh-profile has been a simple proxy of GGA for sometime.


kabuumu:
This code isn't used so we're going to be deleting it. Thank you for pointing this out.

PR #355, opened 2018-01-22
SAR-1272b: re added the tax year deferral behind a feature switch to …

suyangyu:
…accomendate new requirements

### DIFF
--- a/app/incometax/business/controllers/BusinessAccountingMethodController.scala
+++ b/app/incometax/business/controllers/BusinessAccountingMethodController.scala
@@ -20,9 +20,11 @@ import javax.inject.{Inject, Singleton}
 
 import core.auth.SignUpController
 import core.config.BaseControllerConfig
+import core.services.CacheUtil.CacheMapUtil
 import core.services.{AuthService, KeystoreService}
 import incometax.business.forms.{AccountingMethodForm, MatchTaxYearForm}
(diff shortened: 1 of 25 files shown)


app/incometax/business/controllers/BusinessAccountingMethodController.scala
```diff
@@ -73,12 +76,31 @@ class BusinessAccountingMethodController @Inject()(val baseConfig: BaseControlle
...
     if (isEditMode)
       Future.successful(incometax.subscription.controllers.routes.CheckYourAnswersController.show().url)
-    else
-      keystoreService.fetchMatchTaxYear() map {
+    else if (applicationConfig.taxYearDeferralEnabled) {
+      for {
+        cacheMap <- keystoreService.fetchAll().map(_.get)
+        matchTaxYear = cacheMap.getMatchTaxYear()
+        optEndTaxYear = cacheMap.getAccountingPeriodDate().map(_.taxEndYear)
+      } yield (matchTaxYear, optEndTaxYear) match {
```
kabuumu: What if matchTaxYear is None?

suyangyu: it fails... but that's fine cos they should have gone through match tax year and answered it by this point

PR #363, opened 2018-01-25
Sar 1283

suyangyu:


app/incometax/subscription/views/check_your_answers.scala.html
```diff
@@ -170,7 +165,9 @@ <h2 class="heading-medium">@Html(Messages("summary.h2"))</h2>
         @summaryModel.incomeSource.map(incomeSource)
         @summaryModel.rentUkProperty.map(rentUkProperty)
         @summaryModel.rentUkProperty.map(onlySourceOfIncome)
-        @summaryModel.workForYourself.map(WorkForYourself)
+        @if(summaryModel.rentUkProperty.fold(false)(_.needSecondPage)) {
```
kabuumu: Can you use exists instead of fold(false) here please?

PR #474, opened 2019-07-22
Added in not eligible view

Daveo-G:


app/incometax/eligibility/views/NotEligibleForIncomeTax.scala.html
```diff
@@ -0,0 +1,35 @@
+@*
```
ChrisHillCap: could this view be named _ kebab case like all the views to keep it consistent
not_eligible_for..

app/incometax/eligibility/views/NotEligibleForIncomeTax.scala.html
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.play.views.html.helpers.form
+@import core.views.html.helpers._
+@import core.config.AppConfig
+@import core.views.html.templates.main_template
+
+@(postAction: Call)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
```
ChrisHillCap: does this need to be a post action, we can pass it in, but it should be just a getAction? i think? theres no form required

ChrisHillCap: edit: if this parameter is not used then it is not needed, so we can just pass the implicits in and nothing else

Daveo-G: Done

app/incometax/eligibility/views/NotEligibleForIncomeTax.scala.html
Daveo-G: Done

PR #486, opened 2019-09-04
SAR-4503: Updated Terms and CYA controllers & tests to skip Terms if feature switch is enabled

brennaksamuel:


app/incometax/subscription/controllers/CheckYourAnswersController.scala
```diff
@@ -78,7 +91,7 @@ class CheckYourAnswersController @Inject()(val baseConfig: BaseControllerConfig,
     Authenticated.async { implicit request =>
       implicit user =>
         keystoreService.fetchAll().flatMap { cache =>
-          cache.getTerms match {
+          (if (appConfig.eligibilityPagesEnabled) Some(true) else cache.getTerms) match {
```
kabuumu: Could this logic be split out for the new feature switched logic and the old? The terms check will become irrelevant after the feature switch is removed so we want to make it easy to delete.

PR #487, opened 2019-09-09
SAR-4504: Skip other income page when feature switch is enabled

r-melvin:


app/incometax/subscription/views/check_your_answers.scala.html
```diff
@@ -169,7 +169,11 @@ <h2 class="heading-medium">@Html(Messages("summary.h2"))</h2>
         @if(summaryModel.rentUkProperty.exists(_.needSecondPage)) {
             @summaryModel.workForYourself.map(WorkForYourself)
         }
-        @summaryModel.otherIncome.map(otherIncome)
+        @if(appConfig.eligibilityPagesEnabled) {
```
AlexRimmerHMRC: Twirl can handle the absence of an else, so this could be changed to if(!appConfig.eligibilityPagesEnabled)

PR #491, opened 2019-09-17
SAR-4540 Added service for calculating whether accounting period is v…

MarjanMubarok:


app/core/services/AccountingPeriodService.scala
```diff
@@ -0,0 +1,40 @@
...
+package core.services
+
+import java.time.LocalDate
+
+import incometax.util.{AccountingPeriodUtil, CurrentDateProvider}
+import javax.inject.{Inject, Singleton}
+
+@Singleton
+class AccountingPeriodService @Inject()(currentDateProvider: CurrentDateProvider) {
+  def accountingPeriodElligibilty(starDate: LocalDate, endDate: LocalDate): Boolean = {
```
r-melvin: could you name this something like isEligible or isEligiblePeriod

app/core/services/AccountingPeriodService.scala
```diff
@@ -0,0 +1,40 @@
...
+    isWholeYear(starDate, endDate) && isValidTaxYear(starDate, endDate)
+  }
+
+  def isWholeYear(startDate: LocalDate, endDate: LocalDate): Boolean = {
```
r-melvin: Do these methods need to be exposed or can they be private?

app/core/services/AccountingPeriodService.scala
```diff
@@ -0,0 +1,40 @@
...
+    startDate.minusDays(1).plusYears(1) == endDate
+  }
+
+  def isValidTaxYear(startDate: LocalDate, endDate: LocalDate): Boolean = {
+    val currentTaxYearEndYear = AccountingPeriodUtil.getTaxEndYear(currentDateProvider.getCurrentDate())
```
r-melvin: val endOfCurrentTaxYear?

app/core/services/AccountingPeriodService.scala
MarjanMubarok: they are no longer separate and have been moved into the method.

PR #497, opened 2019-09-19
SAR-4581: Amended logic for accounting period service and added test case

brennaksamuel:


app/incometax/util/AccountingPeriodUtil.scala
```diff
@@ -23,9 +23,9 @@ import incometax.business.models.AccountingPeriodModel
 
 
 object AccountingPeriodUtil {
-  private val april = 4
-  private val fifth = 5
-  private val sixth = 6
+  val april = 4
```
kabuumu: These are somewhat misleading, as 4 is april, but 5 and 6 are both days. There's nothing wrong with what the code was previously. Using the standard Java constants is a better idea when they're available.

app/incometax/util/AccountingPeriodUtil.scala
```diff
@@ -35,10 +35,13 @@ object AccountingPeriodUtil {
...
   def getCurrentTaxEndYear: Int = getTaxEndYear(LocalDate.now())
 
-  def getCurrentTaxYearStartDate = DateModel(sixth.toString, april.toString, (getCurrentTaxEndYear -1).toString)
+  def getCurrentTaxYearStartDate = DateModel(sixth.toString, april.toString, (getCurrentTaxEndYear - 1).toString)
 
   def getCurrentTaxYearEndDate = DateModel(fifth.toString, april.toString, getCurrentTaxEndYear.toString)
 
   def getCurrentTaxYear = AccountingPeriodModel(getCurrentTaxYearStartDate, getCurrentTaxYearEndDate)
 
+  def isBetween(date: LocalDate, minDate: LocalDate, maxDate: LocalDate): Boolean =
```
kabuumu: I don't feel like this makes the code any cleaner, it just makes it harder to follow without an IDE.

app/core/services/AccountingPeriodService.scala
```diff
@@ -17,32 +17,30 @@
...
-  private def isEligibleStartDate(startDate: LocalDate): Boolean = {
-    startDate.getMonth == APRIL && startDate.getDayOfMonth >= 1
+  def checkEligibleAccountingPeriod(startDate: LocalDate, endDate: LocalDate): Boolean =
+    if (endDate.getDayOfMonth == 5 && endDate.getMonth == APRIL)
+      validTaxYear(startDate, endDate)
+    else
+      startDate.minusDays(1).plusYears(1) == endDate &&
+        validTaxYear(startDate, endDate)
+
+  private def validTaxYear(startDate: LocalDate, endDate: LocalDate): Boolean = {
```
kabuumu: By calling this in all scenarios, you're performing an additional check that isn't needed for the non April 5th end dates.

PR #500, opened 2019-09-24
SAR-4550 Update the controller to kickout ineligible dates

JSimmons072019:


app/agent/forms/AccountingPeriodDateForm.scala
```diff
@@ -114,4 +114,11 @@ object AccountingPeriodDateForm {
     )(AccountingPeriodModel.apply)(AccountingPeriodModel.unapply).verifying(endDateAfterStart andThen endDate24MonthRule)
   )
 
+  val accountingPeriodDates = Form(
```
david-walters-HMRC: accountingPeriodDateForm, cant this be removed? is it used anywhere

JSimmons072019: Yes it can, forgot to take it out

PR #517, opened 2019-10-21
SAR-4673 - Removed ES8 feature switch and associated code.

MarjanMubarok:


app/incometax/subscription/services/EnrolmentService.scala
```diff
@@ -37,12 +35,8 @@ class EnrolmentService @Inject()(config: AppConfig,
                                  ggConnector: GGConnector,
```
AlexRimmerHMRC: This ggConnector seems to no longer be used, can it be removed, along with anything related

PR #518, opened 2019-10-21
SAR_4672 Removing EMAC ES6 feature switch

JSimmons072019:


test/incometax/subscription/services/mocks/TestKnownFactsService.scala
```diff
@@ -32,20 +32,10 @@ import scala.concurrent.Future
 trait TestKnownFactsService extends MockGGAdminConnector with MockTaxEnrolmentsConnector with MockConfig {
```
AlexRimmerHMRC:
```suggestion
trait TestKnownFactsService extends MockTaxEnrolmentsConnector with MockConfig {
```

AlexRimmerHMRC: GGAdminStub can also be removed from the code now that it is no longer used

AlexRimmerHMRC: GGAdminConnector is also no longer used and so can be removed along with anything else that uses

PR #533, opened 2019-11-21
SAR-4860 Creating the Agent C/A feature switch

JSimmons072019:


app/core/config/featureswitch/FeatureSwitch.scala
```diff
@@ -84,4 +85,9 @@ case object PropertyCashOrAccruals extends FeatureSwitch {
 case object WhatTaxYearToSignUp extends FeatureSwitch {
   override val displayText: String = "Enable what tax year do you want to sign up for"
   override val name: String = s"$prefix.what-tax-year-to-sign-up"
+}
+
+case object AgentPropertyCashOrAccruals extends FeatureSwitch {
+  override val displayText: String = "Enable cash or accruals choice for property income"
+  override val name: String = s"$prefix.property-cash-or-accruals"
```
AlexRimmerHMRC: Can these be changed, the display text is on our feature switch page and looks the same as the non agent version so will cause confusion.
name is used for selecting it in our acceptance tests, so that will need to be different to the non agent version

JSimmons072019: sorry, I've done it now

PR #547, opened 2019-12-10
 SAR-4961:Update flow to include page when F/S is enabled

scullythornton1:


app/agent/controllers/business/BusinessAccountingMethodController.scala
```diff
@@ -80,7 +80,11 @@ class BusinessAccountingMethodController @Inject()(val baseConfig: BaseControlle
   def backUrl(isEditMode: Boolean): String =
     if (isEditMode)
       agent.controllers.routes.CheckYourAnswersController.show().url
-    else
-      agent.controllers.business.routes.BusinessAccountingPeriodDateController.show().url
-
+    else {
+      if (isEnabled(AgentTaxYear)) {
+        agent.controllers.business.routes.MatchTaxYearController.show().url
```
AlexRimmerHMRC: When the feature switch is enabled, the code should check the answer that was given on the MatchTaxYear page, if it was Yes (Match Tax Year) then it will be the MatchTaxYear page's url, if it was No, it will be the BusinessAccountingPeriod page's url

it/agent/controllers/business/BusinessAccountingPeriodPriorControllerISpec.scala
```diff
@@ -105,6 +110,24 @@ class BusinessAccountingPeriodPriorControllerISpec extends ComponentSpecBase {
...
+        enable(AgentTaxYear)
+
+        Given("I setup the Wiremock stubs")
+        AuthStub.stubAuthSuccess()
+        KeystoreStub.stubKeystoreSave(CacheConstants.AccountingPeriodPrior, userInput)
+
+        When("POST /business/accounting-period-prior is called")
+        val res = IncomeTaxSubscriptionFrontend.submitBusinessAccountingPeriodPrior(inEditMode = false, Some(userInput))
+
+        Then("Should return a SEE_OTHER with a redirect location of register next period")
```
AlexRimmerHMRC:
```suggestion
        Then("Should return a SEE_OTHER with a redirect location of match tax year")
```

PR #570, opened 2020-02-05
MISUV-222 Remove the language feature switch and fix broken tests

JSimmons072019:


scullythornton1:
in config , application.config line 181 need to be removed.

PR #571, opened 2020-02-06
MISUV-208: restructure views

scullythornton1:


app/views/templates/govuk_wrapper.scala.html
```diff
@@ -15,88 +15,86 @@
  *@
 
 @(appConfig: core.config.AppConfig,
-        title: String,
```
JSimmons072019: undo the bad formatting please

PR #572, opened 2020-02-07
MISUV-198 Removed guidance feature switch from code.

jameshforster:


JSimmons072019:
You need to look in: 

application.conf
HomeControllerISpec(both agent and user-matching)
it/helpers/ComponentSpecBase (and the agent variation of this file)

and remove the references to the guidance feature switch

app/controllers/agent/HomeController.scala
```diff
@@ -36,13 +36,8 @@ class HomeController @Inject()(override val baseConfig: BaseControllerConfig,
...
                               ) extends StatelessController {
 
-  lazy val showGuidance: Boolean = baseConfig.applicationConfig.showGuidance
-
   def home: Action[AnyContent] = Action.async { implicit request =>
-    showGuidance match {
-      case true => Ok(views.html.agent.agent_frontpage(controllers.agent.routes.HomeController.index()))
-      case _ => Redirect(controllers.agent.routes.HomeController.index())
-    }
+    Ok(views.html.agent.agent_frontpage(controllers.agent.routes.HomeController.index()))
```
AlexRimmerHMRC: Aren't we supposed to be removing this page?

jameshforster: I based this off the current state of production so I'm not sure? I just literally kept the code which corresponded to that. If not then why is this still enabled in prod?

AlexRimmerHMRC: I don't know if it was even known that this was still enabled, prototype doesn't have it, we don't link to the page from the eligibility service, which is also a point, we wouldn't have this after eligibility questions anyways, we already have govuk guidance before our service. I thought the decision was to remove the page because its redundant.

jameshforster: Alright, I will change it over then. Still seems rather bizarre that this was left enabled.

AlexRimmerHMRC: Yeah, might have been something which got left during team ups and downs

PR #593, opened 2020-04-14
MISUV-390 Replace the ExitSurvey with feedback

JSimmons072019:


conf/agent.routes
```diff
@@ -16,6 +16,11 @@ GET         /confirmation                           controllers.agent.Confirmati
 # Session Timeout Route
 GET         /session-timeout                        controllers.agent.SessionTimeoutController.show
 
+# Feedback Routes
+GET         /feedback                               controllers.FeedbackController.show
+POST        /feedback                               controllers.FeedbackController.submit
+GET         /thankyou                               controllers.FeedbackController.thankyou
+
```
AlexRimmerHMRC, line 23: This shouldn't be required, app.routes has all the routes that should be required

PR #601, opened 2020-05-13
MISUV-475: Combine models, keystore, cache utils for individual and a…

scullythornton1:
…gent journeys


test/controllers/utils/RequirementsSpec.scala
```diff
@@ -117,21 +116,10 @@ class RequirementsSpec extends WordSpecLike with MustMatchers with MockitoSugar
   }
 
   "require" must {
-    "redirect to the no data redirect location" when {
```
chandraAndem, line 120: Looks like a valid test case, any specific reason to remove this?

scullythornton1: before fetchAll function for the agent is a type of [Option], after combining the individual and agent KeyStoreServices, the remove the fetchAll for the agent and kept the fetchAll for the individual which is returning the type Future[CacheMap], thus there will not be a scenario when return "No data".

app/controllers/individual/subscription/ConfirmationController.scala
```diff
@@ -18,23 +18,21 @@ package controllers.individual.subscription
...
-import utilities.individual.CacheUtil._
+import utilities.CacheUtil._
 import views.html.individual.incometax.subscription.sign_up_complete
 
 import scala.concurrent.ExecutionContext
 
 @Singleton
 class ConfirmationController @Inject()(val authService: AuthService, keystoreService: KeystoreService)(implicit val ec: ExecutionContext, appConfig: AppConfig,
-                                       mcc: MessagesControllerComponents) extends PostSubmissionController {
+                                                                                                       mcc: MessagesControllerComponents) extends PostSubmissionController {
```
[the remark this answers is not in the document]
scullythornton1: amended. :)

test/controllers/utils/RequirementsSpec.scala
chandraAndem: Thanks!

PR #605, opened 2020-06-03
Misuv 520

juamalBlackman:


app/views/individual/incometax/incomesource/rent_uk_property.scala.html
```diff
@@ -33,6 +33,18 @@
 @scripts = {
     <script type="text/javascript" src="@routes.Assets.at("javascripts/show-hide-content.js")"></script>
 }
+    @rentUkPropertyContent ={
```
jameshforster, line 36: Why is this page being edited as part of this story? This is to be done as part of MISUV-499 and will only cause conflicts.

AlexRimmerHMRC: To add the correct functionality for the conditionally revealing content in the radio buttons correctly, we had to update the radioHelper and as a result, this page had to be updated a little bit. The conflict with the other story will be simple as they will just be removing this page

juamalBlackman:
Fixed typo, added tests and squashed commits

PR #606, opened 2020-06-03
Controller and View changes for Add income source page

chandraAndem:


app/views/individual/incometax/incomesource/income_source.scala.html
```diff
@@ -0,0 +1,84 @@
...
+         tabindex="-1">
+        <h2 id="error-summary-heading" class="h3-heading">@Messages("base.error_heading")</h2>
+        <ul>
+            <li>
+                <a id="error-summary-heading-href" href="#@IncomeSourceForm.business">@{Messages("individual.error.income_source.invalid")}</a>
+            </li>
+        </ul>
+    </div>
+
+    }
```
AlexRimmerHMRC, line 47: Isn't there an errorSummary helper to do this?

app/models/individual/incomesource/IncomeSourceModel.scala
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.individual.incomesource
+
+import play.api.libs.json._
+
+case class IncomeSourceModel(areYouSelfEmployedModel: Boolean, rentUkPropertyModel: Boolean) {
```
AlexRimmerHMRC, line 21: Could you rename the these two?
Doesn't make sense for them to have model in the name.
Maybe selfEmployment and ukProperty?

app/views/individual/incometax/incomesource/income_source.scala.html
chandraAndem: Yes but helper adds href id from the form error key, which we don't have here as we are not overriding bind and throwing form error.

app/models/individual/incomesource/IncomeSourceModel.scala
chandraAndem: Done.

PR #620, opened 2020-07-27
MISUV-666: Update IncomeSource Redirect

AIbrahimQA:


app/controllers/individual/incomesource/IncomeSourceController.scala
```diff
@@ -60,26 +62,28 @@ class IncomeSourceController @Inject()(val authService: AuthService, keystoreSer
               incomeSource match {
                 case IncomeSourceModel(true, false) =>
                   Redirect(controllers.individual.business.routes.BusinessNameController.show())
-                case IncomeSourceModel(false, true) =>
-                  Redirect(controllers.individual.business.routes.PropertyAccountingMethodController.show())
+                case IncomeSourceModel(false, true) if appConfig.isReleaseFourEnabled =>
```
jameshforster, line 65: Need case for only property model while release four is disabled to route to property accounting method (as removed)

e.g. "case IncomeSourceModel(false, true) => Redirect(controllers.individual.business.routes.PropertyAccountingMethodController.show())" without an if statement

jameshforster, line 65: Must use feature switch functionality NOT appconfig as appconfig is static and cannot be edited.

test/controllers/individual/incomesource/IncomeSourceControllerSpec.scala
```diff
@@ -111,6 +111,46 @@ class IncomeSourceControllerSpec extends ControllerBaseSpec
...
+
+              await(goodRequest)
+              verifyKeystoreFetch(IndividualIncomeSource, 0)
+              verifyKeystoreSave(IndividualIncomeSource, 1)
+            }
+          }
+        }
+      }
+
+      "The redirect to Property Accounting page with Release Four enabled" when {
```
AlexRimmerHMRC, line 134: Test description incorrect for test case

PR #626, opened 2020-08-03
Updating incomesource model with foreignUkProperty

chandraAndem:


app/forms/individual/incomesource/IncomeSourceForm.scala
```diff
@@ -24,11 +24,13 @@ object IncomeSourceForm {
...
   val business = "Business"
   val ukProperty = "UkProperty"
+  val foreignUkProperty = "ForeignUkProperty"
 
   val incomeSourceForm: Form[IncomeSourceModel] = Form(
     mapping(
       business -> boolean,
-      ukProperty -> boolean
+      ukProperty -> boolean,
+      foreignUkProperty -> boolean
```
jameshforster, line 33: should this be foreignProperty instead of foreignUkProperty?

app/models/individual/incomesource/IncomeSourceModel.scala
```diff
@@ -18,8 +18,10 @@ package models.individual.incomesource
 
 import play.api.libs.json._
 
-case class IncomeSourceModel(selfEmployment: Boolean, ukProperty: Boolean) {
-  val hasAtLeastOneSelected: Boolean = selfEmployment || ukProperty
+case class IncomeSourceModel(selfEmployment: Boolean,
+                             ukProperty: Boolean,
+                             foreignUkProperty: Boolean) {
```
jameshforster, line 23: See above comment on naming

app/forms/individual/incomesource/IncomeSourceForm.scala
chandraAndem: Don't know what was running in my mind :face_palm: 
Changed now.

app/models/individual/incomesource/IncomeSourceModel.scala
chandraAndem: Done

PR #642, opened 2020-08-24
MISUV-720: Update links

AIbrahimQA:


app/controllers/individual/business/BusinessAccountingMethodController.scala
```diff
@@ -67,6 +69,11 @@ class BusinessAccountingMethodController @Inject()(val authService: AuthService,
               subscriptionDetailsService.fetchIndividualIncomeSource() map {
                 case Some(IncomeSourceModel(_, true, _)) =>
                   Redirect(controllers.individual.business.routes.PropertyAccountingMethodController.show())
+                case Some(IncomeSourceModel(_, _, true)) =>
```
jameshforster, line 72: This can be simplified to:
case Some(IncomeSourceModel(_, _, true)) if isEnabled(ForeignProperty) => Redirect(controllers.individual.business.routes.OverseasPropertyCommencementDateController.show())

app/controllers/individual/business/PropertyAccountingMethodController.scala
```diff
@@ -66,8 +66,20 @@ class PropertyAccountingMethodController @Inject()(val authService: AuthService,
...
           view(accountingMethodPropertyForm = formWithErrors, isEditMode = isEditMode).map(view => BadRequest(view)),
         accountingMethodProperty => {
-          subscriptionDetailsService.saveAccountingMethodProperty(accountingMethodProperty) map { _ =>
-            Redirect(controllers.individual.subscription.routes.CheckYourAnswersController.show())
+          subscriptionDetailsService.saveAccountingMethodProperty(accountingMethodProperty) flatMap { _ =>
+            if (isEditMode) {
+              Future.successful(Redirect(controllers.individual.subscription.routes.CheckYourAnswersController.show()))
+            } else {
+              subscriptionDetailsService.fetchIndividualIncomeSource() map {
+                case Some(IncomeSourceModel(_, _, true)) =>
```
jameshforster, line 74: See above comment

jameshforster (changes requested):
Also requires could you set it so that the foreign property field does not appear on the view? It is appearing regardless of the feature switch at this point.

app/controllers/individual/business/BusinessAccountingMethodController.scala
AIbrahimQA: i was matching the way it was done in Incomesource Controller

PR #659, opened 2020-09-21
Integrating self employments to main flow

chandraAndem:


app/controllers/individual/incomesource/IncomeSourceController.scala
```diff
@@ -57,34 +57,37 @@ class IncomeSourceController @Inject()(val authService: AuthService, subscriptio
...
         }
       )
   }
 
+
+  private def linearJourney(incomeSource: IncomeSourceModel)(implicit request: Request[_]): Future[Result] = {
+    subscriptionDetailsService.saveIncomeSource(incomeSource) map { _ =>
+      incomeSource match {
+        case IncomeSourceModel(true, _, _) =>
+          if (isEnabled(ReleaseFour)) Redirect(controllers.individual.business.routes.WhatYearToSignUpController.show())
```
jameshforster, line 77: This cannot be redirected to unless both property values are false, otherwise this page must be skipped.

app/controllers/individual/incomesource/IncomeSourceController.scala
```diff
@@ -57,34 +57,37 @@ class IncomeSourceController @Inject()(val authService: AuthService, subscriptio
...
+          else Redirect(controllers.individual.business.routes.BusinessNameController.show())
+        case IncomeSourceModel(_, true, _) =>
+          if (isEnabled(ReleaseFour)) Redirect(controllers.individual.business.routes.WhatYearToSignUpController.show())
```
jameshforster, line 80: This is not enabled for either property field, this should redirect to PropertyAccountingMethod only

app/controllers/individual/incomesource/IncomeSourceController.scala
```diff
@@ -57,34 +57,37 @@ class IncomeSourceController @Inject()(val authService: AuthService, subscriptio
...
+          else Redirect(controllers.individual.business.routes.PropertyAccountingMethodController.show())
+        case IncomeSourceModel(_, _, true) =>
+          if (isEnabled(ForeignProperty)) Redirect(controllers.individual.business.routes.WhatYearToSignUpController.show())
```
jameshforster, line 83: As above but for foreign property

jameshforster (changes requested):

app/controllers/individual/incomesource/IncomeSourceController.scala
chandraAndem: Done

app/controllers/individual/incomesource/IncomeSourceController.scala
chandraAndem: Done

app/controllers/individual/incomesource/IncomeSourceController.scala
chandraAndem: Done

PR #675, opened 2020-10-19
MISUV-991 Update flow for Agent UK Property

Wright93:


scullythornton1:
PropertyCommentcementDateController the backUrl will need to update. 

I went through the journey with Release 4 enabled and when I selected both business and property income sources and when I was on the "property commencement date" page click back link button, it led me to "what tax year to sign up" when it actually should lead back to the "business accounting method" page.

the "property commencement date" page will never lead to "what year to sign up" page, as long as the user selects either property income or both business and property, there will be no option for the user to select the tax year. ( the page will not display) as currently we do not allow the property user to select next tax year.

scullythornton1:
sorry to be a pain, but can you please also update the test below:

BusinessNameControllerSepc 

PropertyCommencementDateContollerSpec 
line 190 ( change text:  what year to sign up to business accounting method) 
can you also tidy up the scenarios that related to -->not edit mode, property only, but FS enabled and disabled tests.

PR #687, opened 2020-10-26
MISUV-1059: HTML Validation Error

AIbrahimQA:


app/views/individual/incometax/business/overseas_property_accounting_method.scala.html
```diff
@@ -60,18 +64,8 @@ <h1 class="heading-large">@question</h1>
...
-                            <div>{messages("overseas.property.accounting_method.radio.cash.detail")}</div>
-                    }.mkString
-                ),
-                RadioOption(
-                    AccountingMethodMapping.option_accruals, {
-                            <div class="bold">{messages("overseas.property.accounting_method.radio.standard")}</div>
-                            <div>{messages("overseas.property.accounting_method.radio.standard.detail")}</div>
-                    }.mkString
-                )
+                RadioOption(AccountingMethodMapping.option_cash, Messages("overseas.property.accounting_method.cash")),
```
AlexRimmerHMRC, line 67: Any chance we could keep html out of the messages file? Keeping it similar to how it is done previously?

AIbrahimQA: The error was relating to how it was done before with Div being inside the label. This was the only way that worked for me that kept the same styling with the radio options/hints. Do you have any suggestions?

PR #690, opened 2020-10-28
MISUV-1006 Update CYA with overseas pages and change accruals -> stan…

JSimmons072019:
…dard and basis -> accounting in message files

### DIFF
--- a/app/models/SummaryModel.scala
+++ b/app/models/SummaryModel.scala
@@ -35,6 +35,10 @@ sealed trait SummaryModel {
   def propertyCommencementDate: Option[PropertyCommencementDateModel]
 
   def accountingMethodProperty: Option[AccountingMethodPropertyModel]
+
+  def overseasPropertyCommencementDate: Option[OverseasPropertyCommencementDateModel]
+
+  def overseasAccountingMethodProperty: Option[OverseasAccountingMethodPropertyModel]
 }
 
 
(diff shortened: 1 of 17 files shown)


app/models/SummaryModel.scala
```diff
@@ -84,6 +88,9 @@ case class IndividualSummary(incomeSource: Option[IncomeSourceModel] = None,
     )
   }
 
+  override def overseasPropertyCommencementDate: Option[OverseasPropertyCommencementDateModel] = ???
+
+  override def overseasAccountingMethodProperty: Option[OverseasAccountingMethodPropertyModel] = ???
```
AlexRimmerHMRC, line 93: What are you doing here? IndividualSummary already has these values defined, they just have a different name
`overseasPropertyCommencementDateModel`
`overseasAccountingMethodPropertyModel`
I would just change the name in the case class to remove Model

app/views/agent/helpers/SummaryHelper.scala
```diff
@@ -33,6 +33,11 @@ object SummaryHelper {
     case Accruals => Messages("agent.summary.income_type.accruals")
   }
 
+  def accountingMethodText(src: OverseasAccountingMethodPropertyModel)(implicit messages: Messages): String = src.overseasPropertyAccountingMethod match {
+    case Cash => Messages("agent.summary.income_type.cash")
+    case Accruals => Messages("agent.summary.income_type.accruals")
+  }
```
AlexRimmerHMRC, line 39: Because this function is the exact same as the previous two just with a different model, would it make more sense to instead combine this functions into a single function which just takes an AccountingMethod type?

test/agent/assets/MessageLookup.scala
```diff
@@ -189,7 +189,10 @@ object MessageLookup {
     object IncomeSource {
       val business = "Sole trader business"
       val property = "Property"
-      val both = "Sole trader business and property"
+      val overseas_property = "Overseas property"
+      val business_property = "Sole trader business and property"
+      val business_overseas_property = "Sole trader business dand overseas property"
```
Wright93, line 194: Should this be "and" not "dand"?

PR #695, opened 2020-11-05
MISUV-1060. Updated routes in IncomeSourceController, CYA view and tests

WozCapGem:
…ests

### DIFF
--- a/app/controllers/individual/incomesource/IncomeSourceController.scala
+++ b/app/controllers/individual/incomesource/IncomeSourceController.scala
@@ -18,7 +18,7 @@ package controllers.individual.incomesource
 
 import auth.individual.SignUpController
 import config.AppConfig
-import config.featureswitch.FeatureSwitch.{ForeignProperty, ReleaseFour}
+import config.featureswitch.FeatureSwitch.{ForeignProperty, PropertyNextTaxYear, ReleaseFour}
 import config.featureswitch.FeatureSwitching
 import forms.individual.incomesource.IncomeSourceForm
 import javax.inject.{Inject, Singleton}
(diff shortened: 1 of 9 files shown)


app/controllers/individual/incomesource/IncomeSourceController.scala
```diff
@@ -75,15 +76,19 @@ class IncomeSourceController @Inject()(val authService: AuthService, subscriptio
       incomeSource match {
         case IncomeSourceModel(true, false, false) =>
           if (isEnabled(ReleaseFour)) Redirect(controllers.individual.business.routes.WhatYearToSignUpController.show())
+          else if (isEnabled(PropertyNextTaxYear)) Redirect(controllers.individual.business.routes.WhatYearToSignUpController.show())
```
AlexRimmerHMRC, line 79: This else if shouldn't be here, it should just have the two cases

test/views/individual/incometax/subscription/CheckYourAnswersViewSpec.scala
```diff
@@ -79,13 +79,14 @@ class CheckYourAnswersViewSpec extends UnitTestTrait with ImplicitDateFormatter
   lazy val postAction: Call = controllers.individual.subscription.routes.CheckYourAnswersController.submit()
   lazy val backUrl: String = controllers.individual.subscription.routes.CheckYourAnswersController.show().url
 
-  def page(testSummaryModel: IndividualSummary, releaseFour: Boolean = false): HtmlFormat.Appendable =
+  def page(testSummaryModel: IndividualSummary, releaseFour: Boolean = false, propertyNextTaxYear: Boolean = false): HtmlFormat.Appendable =
```
AlexRimmerHMRC, line 82: along with this change, there should be a test that when this propertyNextTaxYear is true, and the user has property income, the change link shows

PR #704, opened 2020-11-13
MISUV-1098 - Updating Agent Sign Up Complete to include dynamic dates

Wright93:


app/services/AccountingPeriodService.scala
```diff
@@ -46,4 +46,26 @@ class AccountingPeriodService @Inject()(currentDateProvider: CurrentDateProvider
   }
 
   def currentTaxYear: Int = AccountingPeriodUtil.getTaxEndYear(currentDateProvider.getCurrentDate)
+
+  object agentUpdateDates {
```
AlexRimmerHMRC, line 50:
```suggestion
  object AgentUpdateDates {
```
Objects tend to start uppercase

app/utilities/UserMatchingSessionUtil.scala
```diff
@@ -50,6 +50,20 @@ object UserMatchingSessionUtil {
         dobY,
         nino
       )
+
+    def clearUserDetailsExpectName(implicit request: Request[AnyContent]): Result =
```
AlexRimmerHMRC, line 54:
```suggestion
    def clearUserDetailsExcectName(implicit request: Request[AnyContent]): Result =
```
Is this supposed to be Except rather than Expect?

public/stylesheets/mtd-itsa-custom.css
```diff
@@ -242,3 +242,7 @@ input[type=radio]
 .js-enabled #global-header .header-proposition #proposition-links {
     display: block;
 }
+
+.govuk-box-highlight {
+    background: #00703c
+}
```
AlexRimmerHMRC, line 248: Could you double check what the individual `sign up complete` page is using for it's box, as I believe it is already the correct colour so we might not need this new css

PR #712, opened 2020-11-26
Fixed LocalAuthorityGrants id and url. Fixed Eat-Out scheme url. Fixe…

WozCapGem:
…d Welsh message.


scullythornton1:
(Covid19ClaimCheckControllerISpec.scala:69) the test is failed due to the message does not reflect the latest message.

PR #736, opened 2021-01-14
added and implemented agent_main_template to the service

WozCapGem:


AlexRimmerHMRC:
Took a quick look, I think there are quite a few agent views which haven't been swapped over to use the agent_main_template created

PR #763, opened 2021-05-11
MTDFITSASU-128: Add quotes to any attribute value

scullythornton1:


app/views/templates/injected/GovUKWrapper.scala.html
```diff
@@ -69,7 +69,7 @@
 
 @signOutLink = {
 @if(showSignOutLink) {
-    <li><a id="logOutNavHref" href=@signOut>@Messages("base.sign-out")</a></li>
+    <li><a id="logOutNavHref" href="@signOut>@Messages("base.sign-out")"</a></li>
```
AlexRimmerHMRC, line 72:
```suggestion
    <li><a id="logOutNavHref" href="@signOut">@Messages("base.sign-out")</a></li>
```

scullythornton1: done

PR #767, opened 2021-05-19
Added tests for agent

abiodxn:


test/views/ViewSpecTrait.scala
```diff
@@ -442,6 +442,7 @@ trait ViewSpecTrait extends UnitTestTrait {
     }
 
     if(isAgent) {
+      System.out.println("THIS IS THE AGENT")
```
AlexRimmerHMRC, line 445: Please remove println

PR #779, opened 2021-06-17
Create new govuk wrapper and templates

llmikeyj:


app/views/templates/agent_main_template.scala.html
```diff
@@ -63,7 +63,7 @@
     }
 
 @headStyle = {
-        <link rel="stylesheet" href="@routes.Assets.at("stylesheets/banner-panel.css")">
+        <link rel="stylesheet" href="@routes.Assets.versioned("stylesheets/banner-panel.css")">
 }
 
 @prefixedTitle = @{
```
abiodxn, line 69: Do we still need this? We have AgentMainTemplate.scala ?

app/views/templates/injected/MainTemplate.scala.html
```diff
@@ -48,16 +48,16 @@
...
         @* append custom scripts *@
 @scriptElem.fold(HtmlFormat.empty)(x => x)
 }
 
 @headStyle = {
-        <link rel="stylesheet" href="@routes.Assets.at("stylesheets/banner-panel.css")">
+        <link rel="stylesheet" href="@routes.Assets.versioned("stylesheets/banner-panel.css")">
 }
 
 @prefixedTitle = @{
```
abiodxn, line 63: Is this used anywhere? We have main template, and principal main template ?

app/views/templates/main_template.scala.html
```diff
@@ -64,7 +64,7 @@
 }
 
 @headStyle = {
-        <link rel="stylesheet" href="@routes.Assets.at("stylesheets/banner-panel.css")">
+        <link rel="stylesheet" href="@routes.Assets.versioned("stylesheets/banner-panel.css")">
 }
 
 @prefixedTitle = @{
```
abiodxn, line 70: Do we need both main_template and MainTemplate ?

app/views/templates/agent_main_template.scala.html
llmikeyj: I think we do; this ticket is not about decommissioning the old templates, and the changes to the generated ReverseRoutes are breaking changes

app/views/templates/injected/MainTemplate.scala.html
llmikeyj: This appears to have been left over from some previous experiment. Advice from Alex was to leave it for now, in anticipation of a proper clean-up once we've migrated over to the new design system.

app/views/templates/main_template.scala.html
llmikeyj: I think we do, as per previous comments

PR #791, opened 2021-07-06
New SPS connector

llmikeyj:


app/connectors/SPSConnector.scala
```diff
@@ -0,0 +1,38 @@
...
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class SPSConnector @Inject()(appConfig: AppConfig,
+                             http: HttpClient)
+                            (implicit ec: ExecutionContext) {
+
+  def postSpsConfirm(entityId: String, itsaId: String)(implicit hc: HeaderCarrier): Future[HttpResponse] = {
+    http.POSTEmpty[HttpResponse](appConfig.microServiceUrl + s"/channel-preferences/confirm/$entityId/itsa/$itsaId")
```
AlexRimmerHMRC, line 31: microServiceUrl is a reference to the baseUrl used by the backend, locally this becomes: localhost:9560, but we actually want to make a call out to a different port, likely a new app config value should be added for channel-preferences

llmikeyj: Okay

PR #798, opened 2021-07-12
integrate sps connector

llmikeyj:


test/services/SPSServiceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      val service = new SPSService(mockConnector)
+      service.enable(SPSEnabled)
+      val mockRequest = mock[Request[AnyContent]]
+      val mockSession = mock[Session]
+      when(mockSession.get(SPSEntityId)).thenReturn("something")
+      when(mockSession.get(mtdItsaEnrolmentIdentifierKey)).thenReturn("something else")
+      when(mockRequest.session).thenReturn(mockSession)
+      service.confirmPreferences(mockRequest)
+
+      verify(mockConnector).postSpsConfirm(any(), startsWith("HMRC-MTD-IT~MTDITID~"))(any())
```
AlexRimmerHMRC, line 43:
```suggestion
      verify(mockConnector).postSpsConfirm(any(), ArgumentMatchers.eq("HMRC-MTD-IT~MTDITID~something"))(any())
```

app/controllers/individual/subscription/ConfirmationController.scala
```diff
@@ -45,6 +47,7 @@ class ConfirmationController @Inject()(val auditingService: AuditingService,
       val endYearOfCurrentTaxPeriod = accountingPeriodService.currentTaxYear
       val updatesAfter = accountingPeriodService.updateDatesAfter()
       val updatesBefore = accountingPeriodService.updateDatesBefore()
+      spsService.confirmPreferences(request)
```
AlexRimmerHMRC, line 50: I don't think we should be confirming preferences here, this will be called every time the user hits this page, refreshes will therefore result in multiple calls made to the sps service

app/services/SPSService.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.concurrent.ExecutionContext
+
+class SPSService @Inject()(val spsConnector: SPSConnector)
+                          (implicit ec: ExecutionContext) extends FeatureSwitching {
+
+  def confirmPreferences(request: Request[AnyContent])(implicit hc: HeaderCarrier): Unit = {
+    if (isEnabled(SPSEnabled)) {
+      request.session.get(SPSEntityId)
+        .map(entityId => request.session.get(mtdItsaEnrolmentIdentifierKey)
+          .map(itsaId => spsConnector.postSpsConfirm(entityId, s"HMRC-MTD-IT~MTDITID~$itsaId")))
```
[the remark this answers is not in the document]
AlexRimmerHMRC: I don't see where we are getting this mtd id from as the session key was added by this story

PR #800, opened 2021-07-26
ITSASU-fixYearValidation: change max and min start date to def

scullythornton1:


llmikeyj (approved):
LGTM but I would like an explanation :)

PR #802, opened 2021-08-02
ITSASU-297

pradeepkjobs:
COMPLETED TESTS AND DEVELOPEMENT OF THSI TICKET


app/views/agent/CheckYourAnswers.scala.html
```diff
@@ -0,0 +1,196 @@
...
+
+@agentMainTemplate(title = Messages("agent.summary.title"), backLink = backUrl) {
+
+
+<h1 class="govuk-heading-l">@Messages("agent.summary.heading") <span class="govuk-visually-hidden">@Messages("agent.summary.heading_hidden")</span>
+</h1>
+
+<div class="govuk-form-group">
+
+    @if(releaseFour) {
```
llmikeyj, line 166: I think we can remove this feature switch now?

llmikeyj (changes requested):
Looks fine, I believe we should remove references to feature switch 'release four' though

app/views/agent/CheckYourAnswers.scala.html
pradeepkjobs: After discussing with Scully the feature switch will involve much of Controller logic change and tests, it which will be bigger than this story point assigned/scope of the story, this will be covered by the story (ITSASU-160) that removes releasefour

PR #819, opened 2021-09-03
ITSASU-348: migrate already signed up

llmikeyj:


app/views/individual/incometax/subscription/enrolled/AlreadyEnrolled.scala.html
```diff
@@ -0,0 +1,33 @@
...
+@import views.html.templates.PrincipalMainTemplate
+
+@this(
+        mainTemplate: PrincipalMainTemplate,
+        signOutButton: SignOutButton
+)
+
+@()(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+
+@mainTemplate(title = Messages("already-enrolled.title"), showSignOutLink = false) {
```
AlexRimmerHMRC, line 27: Could we remove the showSignOutLink = false, I believe this might have been an issue with the previous page, if we give them the option to sign out via the @signOutButton(), they are signed up, therefore we should have the sign out link

llmikeyj: Does that mean we have two means on the page of doing the same thing though?

AlexRimmerHMRC: Yeah, but design convention is to have the sign out link in the heading banner in the top right, future usability stories will likely address the "2 sign out links" issue

llmikeyj: Okay!

PR #824, opened 2021-09-06
ITSASU-344-mergercommit

pradeepkjobs:


app/views/individual/usermatching/CheckYourUserDetails.scala.html
```diff
@@ -0,0 +1,105 @@
...
+<h1 class="heading-large">@Messages("user-details.summary.heading")</h1>
+
+<dl class="govuk-summary-list">
+    @firstName(userDetailsModel)
+    @lastName(userDetailsModel)
+    @nino(userDetailsModel)
+    @dob(userDetailsModel)
+</dl>
+
+@form(action = postAction) {
```
AlexRimmerHMRC, line 101: Wrong form is being used here, an injected: `form: FormWithCSRF` should be used instead

app/views/individual/usermatching/CheckYourUserDetails.scala.html
```diff
@@ -0,0 +1,105 @@
...
+val rowQuestion = Messages("user-details.summary.dob")
+val rowAnswer = userDetailsModel.dateOfBirth.toCheckYourAnswersDateFormat
+val rowUrl = controllers.usermatching.routes.UserDetailsController.show(editMode = true).url
+val hiddenContent = Some(Messages("userdetail.cya.change.dateofbirth"))
+summaryRow(rowName, rowQuestion, rowAnswer, rowUrl, hiddenContent)
+}
+
+@principalMainTemplate(title = messages("summary.title"), backLink = backUrl) {
+
+<h1 class="heading-large">@Messages("user-details.summary.heading")</h1>
```
AlexRimmerHMRC, line 92: Wrong class used, new heading class should be "govuk-heading-l"

pradeepkjobs: @AlexRimmerHMRC Updated as per above suggestion

PR #827, opened 2021-09-09
Itsau-392, Added Tasklist feature switch

WozCapGem:


app/config/featureswitch/FeatureSwitch.scala
```diff
@@ -30,7 +30,8 @@ object FeatureSwitch {
     IdentityVerification,
     RemoveCovidPages,
     SPSEnabled,
-    ClaimEnrolment
+    ClaimEnrolment,
+    TaskList
```
AlexRimmerHMRC, line 34: TaskList doesn't exist

PR #829, opened 2021-09-15
ITSASU-396, Added TaskListModel and updated tests

WozCapGem:


app/models/common/TaskListModel.scala
```diff
@@ -0,0 +1,49 @@
...
+  var selfEmploymentsComplete: Boolean = (if(selfEmployments.forall(_.isComplete))true else false)
+
+  val overseasPropertyComplete: Boolean = overseasPropertyStart.isDefined && overseasPropertyAccountingMethod.isDefined
+
+  val sectionsTotal: Int = 1 + selfEmployments.size + ukPropertyStart.size + overseasPropertyStart.size
+
+  val sectionsComplete: Int = taxYearSelection.size + selfEmployments.filter(_.isComplete).size + (if(ukPropertyComplete) 1 else 0) +
+    (if(overseasPropertyComplete)1 else 0)
+
+  var canAddMoreBusinesses: Boolean = (if(selfEmployments.size < 50)true else false)
```
AlexRimmerHMRC, line 41: This is slightly incorrect, if the user added 50 self employment businesses, this would mean they could not add any uk property or overseas property.
```suggestion
  var canAddMoreBusinesses: Boolean = if(selfEmployments.size < 50 || ukPropertyStart.isEmpty || overseasPropertyStart.isEmpty) true false
```
Might work better

app/models/common/TaskListModel.scala
```diff
@@ -0,0 +1,49 @@
...
+case class TaskListModel(taxYearSelection: Option[AccountingYear],
+                         selfEmployments: Seq[SelfEmploymentData],
+                         ukPropertyStart: Option[DateModel],
+                         ukPropertyAccountingMethod: Option[AccountingMethod],
+                         overseasPropertyStart: Option[DateModel],
+                         overseasPropertyAccountingMethod: Option[AccountingMethod]) {
+
+  val ukPropertyComplete: Boolean = ukPropertyStart.isDefined && ukPropertyAccountingMethod.isDefined
+
+  var selfEmploymentsComplete: Boolean = (if(selfEmployments.forall(_.isComplete))true else false)
```
AlexRimmerHMRC, line 32: I don't think we have any use for a overarching self employments complete value, we only ever display the complete status of individual businesses and property incomes, so we can likely remove this.
FYI, forall returns a boolean value itself so the if statement isn't actually doing anything
```suggestion
  var selfEmploymentsComplete: Boolean = selfEmployments.forall(_.isComplete)
```
Would do the same thing

app/models/common/TaskListModel.scala
```diff
@@ -0,0 +1,49 @@
...
+  val overseasPropertyComplete: Boolean = overseasPropertyStart.isDefined && overseasPropertyAccountingMethod.isDefined
+
+  val sectionsTotal: Int = 1 + selfEmployments.size + ukPropertyStart.size + overseasPropertyStart.size
+
+  val sectionsComplete: Int = taxYearSelection.size + selfEmployments.filter(_.isComplete).size + (if(ukPropertyComplete) 1 else 0) +
+    (if(overseasPropertyComplete)1 else 0)
+
+  var canAddMoreBusinesses: Boolean = selfEmployments.size < 50  || ukPropertyStart.isEmpty || overseasPropertyStart.isEmpty
+
+  def taskListComplete: Boolean = sectionsComplete == sectionsTotal
```
AlexRimmerHMRC, line 43: Can you add a test for the taskListComplete attribute?

app/models/common/TaskListModel.scala
```diff
@@ -0,0 +1,49 @@
...
+  var selfEmploymentsComplete: Boolean = selfEmployments.forall(_.isComplete)
+
+  val overseasPropertyComplete: Boolean = overseasPropertyStart.isDefined && overseasPropertyAccountingMethod.isDefined
+
+  val sectionsTotal: Int = 1 + selfEmployments.size + ukPropertyStart.size + overseasPropertyStart.size
+
+  val sectionsComplete: Int = taxYearSelection.size + selfEmployments.filter(_.isComplete).size + (if(ukPropertyComplete) 1 else 0) +
+    (if(overseasPropertyComplete)1 else 0)
+
+  var canAddMoreBusinesses: Boolean = selfEmployments.size < 50  || ukPropertyStart.isEmpty || overseasPropertyStart.isEmpty
```
AlexRimmerHMRC, line 41: Can you add a couple tests for this attribute, ie: testing returned values when:
- (less than 50 self employments, uk property started, overseas property started) should be true
- (50 businesses, uk property not started, overseas property started) should be true
- (50 businesses, uk property started, overseas property not started) should be true
- (50 businesses, uk property started + overseas property started) should be false

app/models/common/TaskListModel.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  val ukPropertyComplete: Boolean = ukPropertyStart.isDefined && ukPropertyAccountingMethod.isDefined
+
+  var selfEmploymentsComplete: Boolean = selfEmployments.forall(_.isComplete)
+
+  val overseasPropertyComplete: Boolean = overseasPropertyStart.isDefined && overseasPropertyAccountingMethod.isDefined
+
+  val sectionsTotal: Int = 1 + selfEmployments.size + ukPropertyStart.size + overseasPropertyStart.size
+
+  val sectionsComplete: Int = taxYearSelection.size + selfEmployments.filter(_.isComplete).size + (if(ukPropertyComplete) 1 else 0) +
```
AlexRimmerHMRC, line 38:
```suggestion
  val sectionsComplete: Int = taxYearSelection.size + selfEmployments.count(_.isComplete) + (if(ukPropertyComplete) 1 else 0) +
```
simplification possible

PR #833, opened 2021-09-22
ITSASU-422: Update accounting year model and usage

llmikeyj:


app/utilities/SubscriptionDataUtil.scala
```diff
@@ -75,8 +75,8 @@ object SubscriptionDataUtil extends FeatureSwitching {
         getSelectedTaxYear.getOrElse(
           throw new InternalServerException("[SubscriptionDataUtil][createIncomeSource] - Could not create the create income sources model due to missing selected tax year")
         ) match {
-          case AccountingYearModel(Next) => getNextTaxYear
-          case AccountingYearModel(Current) => getCurrentTaxYear
+          case AccountingYearModel(Next, _) => getNextTaxYear
+          case AccountingYearModel(Current, _) => getCurrentTaxYear
```
AlexRimmerHMRC, line 79: The function here is used to stop users being able to submit bad data. We can't rely on our frontend view to stop users hitting the POST route for the task list page, so we also need to check data, one piece of that data for here specifically should be that confirmed flag.
```suggestion
          case AccountingYearModel(_, false) => throw new InternalServerException("[SubscriptionDataUtil][createIncomeSources] - Could not create the create income sources model as the user has not confirmed their selected tax year")
          case AccountingYearModel(Next, _) => getNextTaxYear
          case AccountingYearModel(Current, _) => getCurrentTaxYear
```
Will prevent the user bypassing that confirmation

llmikeyj: Okay will do

PR #835, opened 2021-09-27
 ITSASU-424: update tax year section display on task list page

scullythornton1:


app/models/common/TaskListModel.scala
```diff
@@ -36,7 +40,7 @@ case class TaskListModel(taxYearSelection: Option[AccountingYear],
 
   val sectionsTotal: Int = Math.max(2, 1 + selfEmployments.size + ukPropertyStart.size + overseasPropertyStart.size)
 
-  val sectionsComplete: Int = taxYearSelection.size +
+  val sectionsComplete: Int = (if(taxYearSelection.exists(_.confirmed)) 1 else 0) +
```
AlexRimmerHMRC, line 43:
```suggestion
  val sectionsComplete: Int = (if(taxYearSelectedAndConfirmed) 1 else 0) +
```
I know why it was why it was, we simplified later so should have gone back to simplify this too :)

scullythornton1: solved

PR #843, opened 2021-10-05
ITSASU-449: migrate overseas accounting method to use new design library

scullythornton1:


app/views/individual/incometax/business/OverseasPropertyAccountingMethod.scala.html
```diff
@@ -0,0 +1,87 @@
...
+
+@(overseasPropertyAccountingMethodForm: Form[OverseasAccountingMethodPropertyModel], postAction: Call, isEditMode: Boolean, backUrl: String)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+
+@question = @{
+    messages("overseas.property.accounting_method.heading")
+}
+
+@detailsContent = {
+    <p>@Messages("property.accounting_method.accordion.content.para")</p>
+    <ul class="bullets">
```
AlexRimmerHMRC, line 44:
```suggestion
    <ul class="govuk-list govuk-list--bullet">
```
wrong class, should be as suggested. I also saw that the same thing was done on the uk property accounting method page, can that be fixed?

PR #847, opened 2021-10-07
ITSASU-463: add save and retrieve flow and update the relative journey

scullythornton1:


app/controllers/individual/business/PropertyAccountingMethodController.scala
```diff
@@ -102,19 +106,25 @@ class PropertyAccountingMethodController @Inject()(val auditingService: Auditing
       )
   }
 
-  def backUrl(isEditMode: Boolean)(implicit hc: HeaderCarrier): Future[String] =
-    if (isEditMode) {
-      Future.successful(controllers.individual.subscription.routes.CheckYourAnswersController.show().url)
-    } else if (isEnabled(ReleaseFour)) {
+  def backUrl(isEditMode: Boolean)(implicit hc: HeaderCarrier): Future[String] = {
+    if (isEnabled(SaveAndRetrieve)) {
```
AlexRimmerHMRC, line 110: Could we add one more scenario into here where when in edit mode, it will go to the task list page else property start date.
I know that the specifics of that aren't "correct" but it will mean less work for the property check your answers story

app/views/agent/SignUpComplete.scala.html
```diff
@@ -71,7 +71,7 @@ <h2 class="govuk-heading-m">@messages("agent.sign-up-complete.whatNow.heading")<
         <p class = "govuk-body">@messages("agent.sign-up-complete.whatNow.para2") </p>
 
         @form(action = postAction) {
-            @buttonHelper(messages("agent.base.sign_up_another_client"))
+            @buttonHelper(None, messages("agent.base.sign_up_another_client"))
```
AlexRimmerHMRC, line 74:
```suggestion
            @buttonHelper(None, messages("agent.base.sign_up_another_client"))
```
Instead of specifying None in each of these cases, could you remove the None and say buttonText = messages("...") instead?

scullythornton1: done

conf/messages.cy
```diff
@@ -20,6 +20,8 @@ base.no                                                         = Na
 base.go_back                                                    = Ewch yn ôl
 base.userResearch.title                                         = Helpu i wella ein gwasanaeth
 base.userResearch.link                                          = Atebwch ychydig o gwestiynau am eich hunan er mwyn ymuno â’r panel ymchwil
+base.save-and-continue-button                                        = Cadw ac yn eich blaen
+base.save-and-come-back-later                                        = Cadw a dod yn ôl yn nes ymlaen
```
AlexRimmerHMRC, line 24: Are these out of line with the surrounding messages?

PR #849, opened 2021-10-08
ITSASU-466: Overseas Property - Start Date Page

llmikeyj:


conf/messages
```diff
@@ -967,3 +967,7 @@ business.progress-saved.content.paragraph2.part2                = to continue. W
 business.progress-saved.content.sign-up-link                    = return to sign up your business for Making Tax Digital for Income Tax
 business.progress-saved.content.sign-out-link                   = sign out and come back later
 business.progress-saved.content.sign-in-link                    = Government Gateway login
+
+
+#Save and retrieve
+save-and-continue-button                                        = Save and continue
```
sb-dev, line 973: It would be go to move this update just below `base.continue` and have `base.save-and-continue`.

app/controllers/individual/business/OverseasPropertyStartDateController.scala
```diff
@@ -43,40 +44,54 @@ class OverseasPropertyStartDateController @Inject()(val auditingService: Auditin
...
+        require(optOverseasPropertyStartDateAnswer and incomeSourceModelAnswer) {
+          case overseasPropertyStartDate ~ incomeSource =>
+            Future.successful(Ok(view(
+              overseasPropertyStartDateForm = form.fill(overseasPropertyStartDate),
+              isEditMode = isEditMode,
+              incomeSourceModel = Some(incomeSource)
+            )))
+        }
+      } else {
+        require(optOverseasPropertyStartDateAnswer) {
```
sb-dev, line 79: Is worth adding a unit test to cover this branch?

app/controllers/individual/business/OverseasPropertyStartDateController.scala
```diff
@@ -43,44 +44,62 @@ class OverseasPropertyStartDateController @Inject()(val auditingService: Auditin
...
                                                     val languageUtils: LanguageUtils,
-                                                    overseasPropertyStartDate: OverseasPropertyStartDate)
+                                                    val overseasPropertyStartDateView: OverseasPropertyStartDate)
                                                    (implicit val ec: ExecutionContext,
                                                     val appConfig: AppConfig,
                                                     mcc: MessagesControllerComponents)
   extends SignUpController with ImplicitDateFormatter with RequireAnswer with FeatureSwitching {
 
-  def view(overseasPropertyStartDateForm: Form[OverseasPropertyStartDateModel], isEditMode: Boolean, incomeSourceModel: IncomeSourceModel)
+  private lazy val isSaveAndRetrieve: Boolean = isEnabled(SaveAndRetrieve)
```
AlexRimmerHMRC, line 53: I don't think we want this to be a lazy val, we often toggle feature switches in our test environments on the go and being a lazy val would cause this to stay as whatever value it was evaluated to first

PR #872, opened 2021-11-05
ITSASU-551: UK Property - Check your answers Page

sb-dev:
Create UK Property model


app/utilities/SubscriptionDataKeys.scala
```diff
@@ -36,6 +36,7 @@ object SubscriptionDataKeys {
   val SelfEmployments = "SelfEmployments"
   val BusinessAccountingMethod = "BusinessAccountingMethod"
   val OverseasPropertyStartDate = "OverseasPropertyStartDate"
+  val Property = "property"
```
AlexRimmerHMRC, line 39: To keep it consistent with the rest of the values, can we have it as `Property`

PR #875, opened 2021-11-10
ITSASU-534: Create new uk property CYA page

sb-dev:


it/helpers/IntegrationTestModels.scala
```diff
@@ -166,6 +166,9 @@ object IntegrationTestModels {
                        selectedTaxYear: Option[AccountingYearModel] = None,
                        businessName: Option[BusinessNameModel] = None,
                        accountingMethod: Option[AccountingMethodModel] = None,
+                       propertyStartDate: Option[PropertyStartDateModel] = None,
+                       propertyAccountingMethod: Option[AccountingMethodPropertyModel] = None,
+                       ukProperty: Option[PropertyModel] = None,
```
AlexRimmerHMRC, line 171: Hey, I think these should be removed as these values are no longer within the cache map

PR #884, opened 2021-11-26
ITSASU-624 - Updated to use a reference for fetching and saving, with…

AlexRimmerHMRC:
… retrieval of reference if it doesn't exist in session


it/controllers/individual/incomesource/IncomeSourceControllerISpec.scala
```diff
@@ -102,82 +102,6 @@ class IncomeSourceControllerISpec extends ComponentSpecBase with FeatureSwitchin
       }
     }
     "not in edit mode and release four is disabled" when {
```
sb-dev, line 104: It would be good to clean up all mentions of `ReleaseFour`:
 - `beforeEach` still contains `disable(ReleaseFour)`
 - `"not in edit mode and release four is disabled"`
 - ` enable(ReleaseFour)`

AlexRimmerHMRC: Hey, I didn't actually remove all logic in the controller which was using the release four feature switch.
I just removed the logic where release four was disabled and it would redirect through to an old business name page. (I didn't want to update the controllers + spec + ispec for pages which we never use in live, so I removed them)
I will go through and remove anything specific for sole trader income though as I removed all the release four logic for that scenario.

sb-dev (approved):

PR #914, opened 2022-01-06
[ITSASU-613] Add tests for confirmed model: changed button message an…

aug24:
…d no save button


pradeepkjobs:
@aug24 Please squash the commits so that we its a single commit

PR #917, opened 2022-01-10
Itsasu 757: add tax year CYA page agent

scullythornton1:


app/views/agent/business/TaxYearCheckYourAnswers.scala.html
```diff
@@ -0,0 +1,59 @@
...
+  mainTemplate: AgentMainTemplate,
+  checkYourAnswers: CheckYourAnswers,
+  checkYourAnswersRow: CheckYourAnswersRow,
+  continueButton: ContinueButton
+)
+
+@(
+  postAction: Call,
+  viewModel: Option[AccountingYearModel],
+  accountingPeriodService: AccountingPeriodService,
```
AlexRimmerHMRC, line 35: You can potentially save a parameter here by removing accountingPeriodService from being passed into the view and instead, move it up into the @this() injected list.
AccountingPeriodService is a injectable class which can easily be injected directly into the view.
This would also mean that the controller doesn't need to be responsible in providing that service to the view.

app/views/agent/business/TaxYearCheckYourAnswers.scala.html
```diff
@@ -0,0 +1,59 @@
...
+
+@mainTemplate(title = messages("business.check-your-answers.title"), backLink = Some(backUrl)) {
+
+  @checkYourAnswers(heading = messages("business.check-your-answers.title"), postAction) {
+    @checkYourAnswersRow(
+      field = "tax-year",
+      question = messages("business.check-your-answers.content.tax-year.question"),
+      answer = taxYear(viewModel),
+      changeUrl = Some(controllers.individual.business.routes.WhatYearToSignUpController.show(editMode = true).url),
+      hiddenContent = Some(messages("business.check-your-answers.content.tax-year.question"))
```
AlexRimmerHMRC, line 55: Can you look at this hiddenContent message, the change url hidden text should be saying "Change ..." but this one doesn't have the "Change" it just has the same text as the question itself.

PR #939, opened 2022-02-08
[ITSASU-862] Add tests for message equivalence and remove unused mess…

aug24:
…ages

Quick cheeky extra card while I wait for the end of the sprint.


test/config/MessagesSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.scalatest.Matchers.convertToAnyShouldWrapper
+
+import scala.io.Source
+
+class MessagesSpec extends FunSuite {
+
+  lazy val messageKeysEnglish = getMessageKeys("messages")
+  lazy val messageKeysWelsh = getMessageKeys("messages.cy")
+
+  test("No messages present in Welsh, but not in English") {
```
sb-dev, line 29: The message is a bit confusing, I would replace it with `Messages present in Welsh (conf/messages.cy) should also have an English translation (conf/messages)`.

test/config/MessagesSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+    val keysInWelshNotEnglish = messageKeysWelsh -- messageKeysEnglish
+    keysInWelshNotEnglish foreach println
+    keysInWelshNotEnglish.size shouldBe 0
+  }
+
+  test("No messages present in English, but not in Welsh") {
```
sb-dev, line 35: The message is a bit confusing, I would replace it with `Messages present in English (conf/messages) should also have a Welsh translation (conf/messages.cy)`.

PR #944, opened 2022-02-17
Itsasu-865, added connectors for deleting data by reference

WozCapGem:


app/connectors/IncomeTaxSubscriptionConnector.scala
```diff
@@ -48,6 +49,11 @@ class IncomeTaxSubscriptionConnector @Inject()(appConfig: AppConfig,
     http.DELETE[HttpResponse](deleteURL(reference))
   }
 
+  def deleteSubscriptionDetails[T](reference: String, key: String)
```
AlexRimmerHMRC, line 52:
```suggestion
  def deleteSubscriptionDetails(reference: String, key: String)
```
can remove the generic type as it's not used here

PR #953, opened 2022-02-24
[ITSASU-892] Remove release four switch, keeping "true" side of any decisions

aug24:


test/utilities/SubscriptionDataUtilSpec.scala
```diff
@@ -289,9 +288,10 @@ class SubscriptionDataUtilSpec extends UnitTestTrait
         IndividualSummary(
           incomeSource = testIncomeSourceProperty,
           accountingMethodProperty = testAccountingMethodProperty,
-          selectedTaxYear = None
+          selectedTaxYear = testSelectedTaxYearNext
```
aug24, line 291: This was an interesting section.  The tests here were not good for coverage of the 'with release 4' code, and I had to work out what the correct answer should be.

scullythornton1 (commented):
in file application.config, there is still enable-release-four = false need to be removed.

PR #954, opened 2022-02-25
[ITSASU-897] Add prepop to eligibility response model

aug24:
This is our best guess of the correct values for field names and types.  It may be necessary to revisit any and all of these when the spec is complete and available.


app/connectors/individual/eligibility/httpparsers/GetEligibilityStatusHttpParser.scala
```diff
@@ -16,32 +16,21 @@
...
-  private val keyCurrent: String = "eligibleCurrentYear"
-  private val keyNext: String = "eligibleNextYear"
-
   implicit object GetEligibilityStatusHttpReads extends HttpResultParser[EligibilityStatus] {
     override def read(method: String, url: String, response: HttpResponse): HttpResult[EligibilityStatus] = response.status match {
-      case OK => for {
-        currentYear <- parseBoolean(response, keyCurrent)
-        nextYear <- parseBoolean(response, keyNext)
-      } yield EligibilityStatus(currentYear, nextYear)
+      case OK => Json.parse(response.json.toString()).validate[EligibilityStatus] match {
```
AlexRimmerHMRC, line 29: `Json.parse(response.json.toString())`
What is the purpose of this part? Turning it into a string and then back into json?

aug24: Let us never speak of that again.

PR #961, opened 2022-03-09
[ITSASU-966] Handle case where business to be deleted has no name. 

aug24:
Also improve tests - the `hasSignOutLink` value was _always_ overridden to true!


aug24:
Marked do not merge pending confirmation of the Welsh value.

sb-dev (approved):

PR #964, opened 2022-03-14
[ITSASU-945] 

aug24:
Make date helper handle optional max/min
Put the errors in the right fields by reimplementing the order of checks


test/forms/agent/ClientDetailsFormSpec.scala
```diff
@@ -162,14 +162,14 @@ class ClientDetailsFormSpec extends PlaySpec with GuiceOneAppPerSuite {
...
-            errors must contain(FormError(s"$clientDateOfBirth-dateDay", error))
+            errors must contain(FormError(s"$clientDateOfBirth-dateMonth", error))
           }
           "has an invalid year" in {
             val error = s"$dateErrorContext.invalid"
 
             val testInput = setupTestData(dob = DateModel("31", "12", "invalid"))
             val errors = clientDetailsForm.bind(testInput).errors
-            errors must contain(FormError(s"$clientDateOfBirth-dateDay", error))
+            errors must contain(FormError(s"$clientDateOfBirth-dateYear", error))
```
aug24, line 172: As you can see, this goes further than previously: instead of going to the first editable field (better than going to the parent id) this goes to the field which is actually in error.

Most of the rest of the changes in test are the same.

app/forms/formatters/NewDateModelMapping.scala
```diff
@@ -16,123 +16,165 @@
...
-    }
-  }
+  def parse(s: String): Option[Int] = Try {
+    s.toInt
+  }.toOption
 
-  def dateModelFormatter(isAgent: Boolean = false, errorContext: String): Formatter[DateModel] = new Formatter[DateModel] {
+  case class DateModelFormatter(isAgent: Boolean,
+                                errorContext: String,
+                                minDate: Option[LocalDate],
```
aug24, line 38: This is the important change: instead of using Play form validation (which will not handle individual fields in the context of other fields) we pass an optional max/min date into the formatter.

app/forms/formatters/NewDateModelMapping.scala
```diff
@@ -16,123 +16,165 @@
...
+            error(ids.totalDayKey, "start_date.maxStartDate", List(dateFormatter.getOrElse(defaultDateFormatter)(max)))
+          case _ => d
+        }
+      }
+      case x => x
+    }
+
+    private def defaultDateFormatter: LocalDate => String = d => d.toString
+
+    private def validateDate(implicit ids: HtmlIds) =
```
aug24, line 121: This is the important part of the processing.  Sequential validation allows us to test for different errors in an order which allows us to get a more accurate answer.

eg previously we checked the day of the month, but this required us also to check the month and year were sane.  This prevented us from returning the correct error when the day validation failed because the month was silly.

app/forms/formatters/NewDateModelMapping.scala
```diff
@@ -16,123 +16,165 @@
...
+      minDate = minDate,
+      maxDate = maxDate,
+      dateFormatter = dateFormatter)
+  )
+
+  // Encapsulation of field ids.
+  // We should never send people to the field with id "key".  Only to an editable field.
+  // The sole exception to this is the fallback position where all our checks have passed
+  // but the values do not parse to a local date for some reason.
+  case class HtmlIds(key: String) {
```
aug24, line 159: This encapsulation makes it easy to confirm that we don't return just the date $key without a context field.

app/controllers/agent/business/OverseasPropertyStartDateController.scala
```diff
@@ -119,6 +119,6 @@ class OverseasPropertyStartDateController @Inject()(val auditingService: Auditin
   }
 
   def form(implicit request: Request[_]): Form[DateModel] = {
-    overseasPropertyStartDateForm(OverseasPropertyStartDateForm.minStartDate.toLongDate, OverseasPropertyStartDateForm.maxStartDate.toLongDate)
+    overseasPropertyStartDateForm(OverseasPropertyStartDateForm.minStartDate, OverseasPropertyStartDateForm.maxStartDate, d => d.toLongDate)
```
aug24, line 122: In order to pass in an actual date for validation instead of strings, but still allow the date to be displayed in the correct way, a formatter is also provided.

aug24, line 122: In order to pass in an actual date for validation instead of strings, but still allow the date to be displayed in the correct way, a formatter is also provided.

sb-dev (approved):

PR #965, opened 2022-03-14
[ITSASU-954] 

aug24:
Bring README into line and add start script in script directory

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -25,10 +25,8 @@ else
    sm --start ITSA_SUBSC_ALL
 fi
 
-sleep 5
 sm --stop INCOME_TAX_SUBSCRIPTION_FRONTEND
 
-sleep 5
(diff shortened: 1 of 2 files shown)


scripts/start
```diff
@@ -0,0 +1,36 @@
...
+   sm --start INCOME_TAX_SUBSCRIPTION_SELF_EMPLOYED_FRONTEND
+   sm --start INCOME_TAX_SUBSCRIPTION_STUBS
+   sm --start INCOME_TAX_SUBSCRIPTION_ELIGIBILITY
+   sm --start INCOME_TAX_SUBSCRIPTION_ELIGIBILITY_FRONTEND
+   sm --start TAX_ENROLMENTS
+else
+   sm --start ITSA_SUBSC_ALL
+fi
+
+sleep 5
```
sb-dev, line 28: `sleep` is not needed because `sm` doesn't run in the background.

PR #986, opened 2022-04-05
[ITSASU-990] Remove Implicits file and usages.

aug24:
Some Implicits were of use.  These have been moved to a better contextual location.


test/services/mocks/MockSpsService.scala
```diff
@@ -37,7 +35,9 @@ trait MockSpsService extends UnitTestTrait with MockitoSugar with BeforeAndAfter
   }
 
   def verifyConfirmPreferencesPostSpsConfirm(testEntityId: String, testMtditid: String, someCount: Option[Int]): Unit =
-    someCount map (count => verify(mockSpsService, times(count)).confirmPreferences(ArgumentMatchers.eq(testEntityId), ArgumentMatchers.eq(testMtditid))(
+    someCount foreach (count =>
```
sb-dev, line 38: 👍

test/controllers/agent/business/PropertyAccountingMethodControllerSpec.scala
```diff
@@ -46,7 +46,7 @@ class PropertyAccountingMethodControllerSpec extends AgentControllerBaseSpec
   override val controllerName: String = "PropertyAccountingMethod"
   override val authorisedRoutes: Map[String, Action[AnyContent]] = Map()
 
-  val propertyAccountingMethodView = mock[PropertyAccountingMethod]
+  private val propertyAccountingMethodView = mock[PropertyAccountingMethod]
```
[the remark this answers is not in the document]
aug24: It's nothing to do with this change, and I deliberately didn't do that (across several PRs) because I suspect there is an issue in the test and it _should_ be used.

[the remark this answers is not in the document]
sb-dev: It's not used because the mocked view is created in this function, line 51:
```
  private def withController(testCode: PropertyAccountingMethodController => Any): Unit = {
    val propertyAccountingMethodView = mock[PropertyAccountingMethod]

    when(propertyAccountingMethodView(any(), any(), any(), any())(any(), any(), any()))
      .thenReturn(HtmlFormat.empty)

    val controller = new PropertyAccountingMethodController(
      propertyAccountingMethodView,
      mockAuditingService,
      mockAuthService,
      MockSubscriptionDetailsService
    )

    testCode(controller)
  }
```

test/views/ViewSpecTrait.scala
```diff
@@ -86,11 +88,13 @@ trait ViewSpecTrait extends UnitTestTrait {
     // n.b. href must be call-by-name otherwise it may not be evaluated with the correct context-root
     def mustHaveALink(id: String, text: String, href: => String): Unit =
       s"$name have a link with text '$text' pointed to '$href'" in {
-        val link = element.getElementById(id)
-        if (link.isEmpty) fail(s"Unable to locate $id")
-        if (!link.tagName().equals("a")) fail(s"The element with id=$id is not a link")
-        link.attr("href") mustBe href
-        link.text() mustBe text
+        Option(element.getElementById(id)) match {
```
sb-dev, line 91: This might be more readable:
```
        Option(element.getElementById(id)) match {
          case None => fail(s"Unable to locate $id")
          case Some(link) if (!link.tagName().equals("a")) => fail(s"The element with id=$id is not a link")
          case Some(link) =>
            link.attr("href") mustBe href
            link.text() mustBe text
        }
```

test/views/agent/ClientDetailsErrorViewSpec.scala
```diff
@@ -45,7 +45,7 @@ class ClientDetailsErrorViewSpec extends ViewSpecTrait {
 
     form.mustHaveContinueButtonWithText(commonMessages.tryAgain)
 
-    testPage.mustHaveSignOutLinkGovUk(commonMessages.signOut, request.path)
+    testPage.mustHaveSignOutLinkGovUk(commonMessages.signOut, Option(request.path))
```
sb-dev, line 48: Did you mean to add `Some` instead of `Option`? I am asking because request has been create with a path which doesn't look like it can be Null:
```
val request = ViewSpecTrait.viewTestRequest

object ViewSpecTrait {
  // these two constants are used for testing the views
  val testBackUrl = "/test-back-url"
  val testCall = Call("POST", "/test-url")
  val viewTestRequest = FakeRequest("POST", "/test-url")
}
```

sb-dev (approved):

PR #987, opened 2022-04-06
[ITSASU-1077] Remove 'old' DateModelMapping and rename NewDateModelMa…

aug24:
…pping to replace it


app/forms/formatters/DateModelMapping.scala
```diff
@@ -22,106 +22,163 @@ import play.api.data.format.Formatter
 import play.api.data.{FormError, Mapping}
 
 import java.time.LocalDate
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
```
aug24, line 25: git diff has got completely confused here. This is not a changed file.  This is a renamed file where a now-deleted file used to be.

test/forms/formatters/DateModelFormatterSpec.scala
```diff
@@ -24,14 +24,14 @@ import play.api.data.format.Formatter
...
-  def dateModelFormatter(isAgent: Boolean = false): Formatter[DateModel] = DateModelMapping.dateModelFormatter(isAgent = isAgent, errorContext = "test")
+  def dateModelFormatter(isAgent: Boolean = false): Formatter[DateModel] = DateModelMapping.DateModelFormatter(isAgent = isAgent, errorContext = "test", None, None, None)
 
   val bindingKey: String = "testKey"
 
   def inputMap(day: Option[String], month: Option[String], year: Option[String]): Map[String, String] = List(
-    s"$bindingKey.${DateModelMapping.day}" -> day,
-    s"$bindingKey.${DateModelMapping.month}" -> month,
-    s"$bindingKey.${DateModelMapping.year}" -> year
+    s"$bindingKey-${DateModelMapping.day}" -> day,
```
aug24, line 32: Old structure updated, and this test now tests the new date model mapping class instead of the old unused one.

As you can see, some of the tests have had to be changed to reflect the more accurate responses.

PR #1016, opened 2022-05-06
ITSASU-1147: change sign up button to submit and continue

scullythornton1:


conf/messages
```diff
@@ -1115,6 +1115,7 @@ business.task-list.content.section2.remove-sole-trader-business = Remove {0}
 business.task-list.content.section2.remove-uk-property          = Remove UK property business
 business.task-list.content.section2.remove-overseas-property    = Remove overseas property business
 business.task-list.content.section3                             = Sign up
+business.task-list.content.section3.submit_and_continue.button  = Submit and continue
```
sb-dev, line 1118: It might be better to add a new entry like `base.submit-and-continue` instead of `business.task-list.content.section3.submit_and_continue.button` so you can use it for the Individual and Agent pages.

sb-dev (approved):

PR #1039, opened 2022-06-10
[ITSASU-1275] Short circuit when session is already in sign up state.

aug24:


app/auth/individual/BaseFrontendController.scala
```diff
@@ -17,7 +17,6 @@
 package auth.individual
 
 import auth.individual.AuthPredicate._
-import auth.individual.JourneyState.{RequestFunctions, SessionFunctions}
```
aug24, line 20: Removed from here because BaseFrontendController, while in a package labelled `individual` is also used by agent controllers.  It was only used in app/controllers/individual/iv/IVSuccessController.scala anyway.

app/controllers/agent/HomeController.scala
```diff
@@ -40,24 +40,32 @@ class HomeController @Inject()(val auditingService: AuditingService,
...
-              }
-            case (Some(_), _) =>
-              Future.successful(Redirect(controllers.agent.matching.routes.NoSAController.show).removingFromSession(ITSASessionKeys.JourneyStateKey))
-            case _ =>
-              Future.successful(Redirect(controllers.agent.matching.routes.ClientDetailsController.show())
-                .addingToSession(ArnKey -> arn).withJourneyState(AgentUserMatching))
-          }
-        case None =>
-          Future.successful(Redirect(controllers.agent.routes.NotEnrolledAgentServicesController.show))
+      val alreadyInSignUp = request.session.isInState(AgentSignUp)
```
aug24, line 43: The main rewrite: nested case statements replaced with a single, clearer, one.

it/helpers/agent/ComponentSpecBase.scala
```diff
@@ -131,16 +131,19 @@ trait ComponentSpecBase extends AnyWordSpecLike with Matchers with OptionValues
   object IncomeTaxSubscriptionFrontend extends UserMatchingIntegrationRequestSupport {
     val csrfToken: String = UUID.randomUUID().toString
 
-    def defaultCookies(withUTR: Boolean = true): Map[String, String] = {
+    def defaultCookies(withUTR: Boolean = true, withJourneyStateSignUp: Boolean = true): Map[String, String] = {
```
aug24, line 134: Needed to be able to create test classes with no journey state at all.

test/config/MockConfig.scala
```diff
@@ -144,5 +146,6 @@ trait MockConfig extends UnitTestTrait with AppConfig {
 object MockConfig extends MockConfig {
 
   override val configuration: Configuration = mock[Configuration]
+  when(configuration.getOptional(any())(any())).thenReturn(None)
```
aug24, line 149: Because we now do the isEnabled call "up front" in the home controller, ensure that we always return None.

test/controllers/agent/AgentControllerBaseSpec.scala
```diff
@@ -88,6 +88,12 @@ trait AgentControllerBaseSpec extends UnitTestTrait with MockAgentAuthService {
     ITSASessionKeys.UTR -> TestConstants.testUtr
   )
 
+  lazy val agentSignUpRequest = FakeRequest().withSession(
```
aug24, line 91: A request without a journey state.

app/auth/individual/BaseFrontendController.scala
aug24: Might be better to move the BaseFrontendController altogether?  Advise if desired and I'll roll it into this change.

AlexRimmerHMRC (dismissed):

AlexRimmerHMRC (approved):

PR #1044, opened 2022-06-15
ITSASU-1311, remove back link from ClientDetailsCYA

WozCapGem:


srikanthsunkara79 (approved):
Code is fine, and manual test passed

PR #1049, opened 2022-06-20
ITSASU-1302 - Implemented individual end throttle into journey

AlexRimmerHMRC:


test/controllers/individual/business/TaskListControllerSpec.scala
```diff
@@ -75,7 +75,8 @@ class TaskListControllerSpec extends ControllerBaseSpec
     MockSubscriptionDetailsService,
     mockSubscriptionOrchestrationService,
     mockIncomeTaxSubscriptionConnector,
-    mockAuthService
+    mockAuthService,
+    app.injector.instanceOf[ThrottlingService]
```
AlexRimmerHMRC, line 79: Didn't add tests into here, or mock the throttling service due to limitations with Mockito.
As it's a Java based library, we aren't able to mock a function which takes a function as a parameter.
Implemented tests for throttling functionality in ISpec instead.

app/services/ThrottlingService.scala
```diff
@@ -25,20 +28,25 @@ import uk.gov.hmrc.http.HeaderCarrier
...
-      .recoverWith { case _ =>
-        logger.warn(s"Throttle ${throttle.throttleId} has failed, recovering with open=${throttle.failOpen}")
-        Future.successful(throttle.failOpen)
-      }
-      .map {
-        case true => success
-        case false => Future.successful(Redirect(throttle.callOnFail))
-      }
-      .flatten
+    if (isEnabled(Throttle)) {
```
aug24, line 36: I dislike calling the Feature Switch "Throttle", which I think I did!  It clashes with the trait name, which is confusing.  Can we come up with anything better?

app/services/ThrottlingService.scala
```diff
@@ -25,20 +28,25 @@ import uk.gov.hmrc.http.HeaderCarrier
...
+          logger.warn(s"Throttle ${throttle.throttleId} has failed, recovering with open=${throttle.failOpen}")
+          Future.successful(throttle.failOpen)
+        }
+        .map {
+          case true => success
+          case false => Future.successful(Redirect(throttle.callOnFail))
+        }
+        .flatten
+    } else {
+      success
```
aug24, line 48: Two invocations of success isn't great. Could we do something nicer with an or of some kind at the top?

app/services/ThrottlingService.scala
AlexRimmerHMRC: Renamed to ThrottlingFeature to be more clear

app/services/ThrottlingService.scala
AlexRimmerHMRC: Implemented this change, success is called only once now.

aug24 (approved):

PR #1074, opened 2022-08-19
ITSASU-1518 - Updated to handle responses from save results, removed …

AlexRimmerHMRC:
…unrequired saves

### DIFF
--- a/app/controllers/usermatching/HomeController.scala
+++ b/app/controllers/usermatching/HomeController.scala
@@ -73,7 +73,7 @@ class HomeController @Inject()(val auditingService: AuditingService,
                 // New phone, who dis?
                 case None => handleNoSubscriptionFound(utr, java.time.LocalDateTime.now().toString, nino)
               }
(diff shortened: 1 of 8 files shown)


app/controllers/usermatching/HomeController.scala
```diff
@@ -128,16 +128,18 @@ class HomeController @Inject()(val auditingService: AuditingService,
...
       reference =>
         subscriptionDetailsService.saveSubscriptionId(reference, mtditId) map {
-          _ =>
+          case Right(_) =>
             Redirect(controllers.individual.subscription.routes.ClaimSubscriptionController.claim)
               .withJourneyState(SignUp)
               .addingToSession(NINO -> nino)
               .addingToSession(UTR -> utr)
+          case Left(_) =>
+            throw new InternalServerException("[HomeController][claimSubscription] - Could not save subscription id")
```
aug24, line 137: Is it not worth logging whatever is in the Left() ?

```suggestion
          case Left(error) =>
            throw new InternalServerException(s"[HomeController][claimSubscription] - Could not save subscription id: $error")
```

aug24 (approved):

PR #1081, opened 2022-09-14
[ITSASU-1580] Alter task list page to match new designs.

aug24:


test/views/agent/AgentTaskListViewSpec.scala
```diff
@@ -81,12 +83,16 @@ class AgentTaskListViewSpec extends ViewSpec {
   )
 
 
-  def page(taskList: TaskListModel = customTaskListModel()): Html = taskListView(
-    postAction = postAction,
-    viewModel = taskList,
-    clientName = "clientName",
-    clientNino = "clientNino"
-  )(request, implicitly, appConfig)
+  val clientName = Random.alphanumeric.take(10).mkString("")
```
AlexRimmerHMRC, line 86: I might be wrong, but can't we just use .mkString without the brackets or ""
I believe the parameterised version is only for adding a separator between the items.
Same for below?

AlexRimmerHMRC (approved):

PR #1086, opened 2022-09-27
ITSASU-1628: Stop checking throttle on the task list show actions

sb-dev:

### DIFF
--- a/app/controllers/agent/ThrottlingController.scala
+++ b/app/controllers/agent/ThrottlingController.scala
@@ -34,13 +34,13 @@ class ThrottlingController @Inject()(val auditingService: AuditingService,
                                      val appConfig: AppConfig) extends BaseFrontendController {
 
   def start(): Action[AnyContent] = Action.async { implicit request =>
-    Future.successful(Ok(throttleStart(backLink(), controllers.agent.routes.TaskListController.show())))
+    Future.successful(Ok(throttleStart(backLink(), controllers.agent.routes.HomeController.index)))
   }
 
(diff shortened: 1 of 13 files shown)


app/controllers/agent/ThrottlingController.scala
```diff
@@ -34,13 +34,13 @@ class ThrottlingController @Inject()(val auditingService: AuditingService,
...
-    Future.successful(Ok(throttleStart(backLink(), controllers.agent.routes.TaskListController.show())))
+    Future.successful(Ok(throttleStart(backLink(), controllers.agent.routes.HomeController.index)))
   }
 
   def end(): Action[AnyContent] = Action.async { implicit request =>
     Future.successful(Ok(throttleEnd(backLink(), controllers.agent.routes.TaskListController.show())))
   }
 
-  private def backLink() = Some(controllers.agent.routes.HomeController.index.url)
+  private def backLink() =controllers.agent.routes.HomeController.index.url
```
AlexRimmerHMRC, line 44: Hmm, thinking about this, I don't think we should be giving the user an option to go back at all.
Given their journey, we can be sure where they used to be, but sending them back to Home isn't really going "back" it could send them forward if they pass the throttle again.
Might just be better removing the back link entirely.
If you want, we can raise as a separate ticket to do this?

sb-dev: yes, let's do it. `controllers.agent.routes.HomeController.index.url` is also used for the back button on the end journey throttle page, which might not be good.

AlexRimmerHMRC (approved):

PR #1111, opened 2022-11-25
ITSASU-1830: Sign up - Confirmation page - Header content (Current Tax Year)

sb-dev:

### DIFF
--- a/app/auth/individual/IncomeTaxSAUser.scala
+++ b/app/auth/individual/IncomeTaxSAUser.scala
@@ -20,8 +20,6 @@ import common.Constants
 import common.Constants.ITSASessionKeys
 import play.api.mvc.{AnyContent, Request}
 import uk.gov.hmrc.auth.core._
-import IncomeTaxSAUser.fullName
-import IncomeTaxSAUser.spsEntityId
 
 trait IncomeTaxUser {
   val enrolments: Enrolments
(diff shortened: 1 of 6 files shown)


app/auth/individual/IncomeTaxSAUser.scala
```diff
@@ -51,6 +49,9 @@ case class IncomeTaxSAUser(enrolments: Enrolments,
     }
   }
 
+  def fullName(implicit request: Request[AnyContent]): Option[String] =
```
aug24, line 52: What is this?  Why not just use the method on user directly, like sps entity?

app/controllers/individual/subscription/ConfirmationController.scala
```diff
@@ -47,7 +47,13 @@ class ConfirmationController @Inject()(val auditingService: AuditingService,
 
           val view = if (isEnabled(ConfirmationPage)) {
             signUpConfirmation(
-              taxYearSelectionIsNext = taxYearSelectionIsNext
+              taxYearSelectionIsNext = taxYearSelectionIsNext,
+              individualUserName = user
+                .fullName
+                .getOrElse(throw new Exception("[ConfirmationController][show]-could not retrieve individual name from session")),
```
aug24, line 53: I don't think this is correct - we are to treat name as optional.  I think if we don't have it, we just don't display it.

conf/messages
```diff
@@ -384,6 +384,8 @@ sign-up-complete.what-now.submit-annual                         = Submit your an
 ## Sign Up Confirmation page ##
 
 sign-up-confirmation.heading                                         = Sign up complete
+sign-up-confirmation.heading.panel                                 = is signed up for Making Tax Digital for Income Tax for the current year
```
aug24, line 387: Whitespace?!

PR #1115, opened 2022-11-28
ITSASU-1791 Client changed property question

willjshark:


aug24:
Can we have a test for the presence of the new content please?

aug24 (approved):

PR #1118, opened 2022-12-08
ITSASU-1905: Use consistent message keys - second part 

sb-dev:


StefanTagg (approved):
Looks good - Discussed with Samir on slack

PR #1119, opened 2022-12-09
ITSASU-1909 Fixed bug with back button url

willjshark:


app/config/frontendAppConfig.scala
```diff
@@ -125,11 +125,14 @@ trait AppConfig {
   def incomeTaxSelfEmploymentsFrontendBusinessAccountingMethodUrl: String
 
   def incomeTaxSelfEmploymentsFrontendClientInitialiseUrl: String
+  def agentOtherSourcesOfIncomeBackUrl: String
```
sb-dev, line 128: I would call this variable `agentIncomeTaxEligibilityFrontendTermsUrl` because it's the name of the page it's pointing to.
`terms` is the short name of the page; for example, we use it for the controller (https://github.com/hmrc/income-tax-subscription-eligibility-frontend/blob/main/conf/agent.routes#L10).

sb-dev (approved):

PR #1134, opened 2023-01-04
ITSASU-1984: Introduce custom messages test

sb-dev:

### DIFF
--- a/app/forms/agent/AccountingYearForm.scala
+++ b/app/forms/agent/AccountingYearForm.scala
@@ -29,8 +29,8 @@ object AccountingYearForm {
   val accountingYearForm: Form[AccountingYear] = Form(
     single(
       accountingYear -> AccountingYearMapping(
-        errInvalid = Invalid("agent.error.what-year.invalid"),
-        errEmpty = Some(Invalid("agent.error.what-year.empty"))
+        errInvalid = Invalid("agent.error.business.what-year-to-sign-up.invalid"),
+        errEmpty = Some(Invalid("agent.error.business.what-year-to-sign-up.empty"))
       ))
   )
 }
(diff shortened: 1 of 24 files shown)


test/views/individual/incometax/subscription/SignUpConfirmationViewSpec.scala
```diff
@@ -38,7 +38,8 @@ class SignUpConfirmationViewSpec extends ViewSpec {
   val testName = "Lisa Khan"
   val testNino = "QQ123456L"
 
-  def page(selectedTaxYearIsNext: Boolean, userNameMaybe: Option[String]): Html = signUpConfirmation(selectedTaxYearIsNext, userNameMaybe, testNino)
+  def page(selectedTaxYearIsNext: Boolean, userNameMaybe: Option[String]): Html =
+    signUpConfirmation(selectedTaxYearIsNext, userNameMaybe, testNino)(implicitly, wrappedMessages, implicitly)
```
aug24, line 42: This is vaguely annoying.  Can't we force it somehow?

AlexRimmerHMRC: The implicit messages are produced from the MessagesApi.
We could replace the default "app" with one we build ourselves and we configure it to replace the MessagesApi it uses with our own which will return the MessagesWrapper as it's implementation of Messages...
Or something like that? @sb-dev

sb-dev: I made an update. I added `implicit lazy val wrappedMessages: Messages = MessagesWrapper(Lang("en"), messagesApi)` to `ViewSpec.scala` because `Messages` are passed implicitly on the apply function.

aug24 (approved):

PR #1145, opened 2023-01-11
[ITSASU-1998] Do not allow editing of tax year if not permitted.

aug24:


app/views/agent/AgentTaskList.scala.html
```diff
@@ -69,18 +69,26 @@
...
     <span class="app-task-list__task-name">
       {if(viewModel.taxYearSelectedAndConfirmed) {
-      <a href={controllers.agent.routes.TaxYearCheckYourAnswersController.show(editMode=true).url} aria-describedby="section_1_thing_1_status" id="select_tax_year_check_your_answer">
-          {viewModel.taxYearSelection.map(_.accountingYear) match {
-              case Some(Current) => messages(s"${messagePrefix}.content.section1.current-tax-year", (accountingPeriodService.currentTaxYear - 1).toString, accountingPeriodService.currentTaxYear.toString)
-              case Some(Next) => messages(s"${messagePrefix}.content.section1.next-tax-year", accountingPeriodService.currentTaxYear.toString, (accountingPeriodService.currentTaxYear + 1).toString)
-              case None => messages(s"${messagePrefix}.content.section1.action")
-          }}
-      </a>
+        if(!viewModel.permitTaxYearChange)
```
sb-dev, line 85: `taxYearSelectedAndConfirmed` is going to return `false` so `taxYearText` needs to be moved somewhere else.

app/models/common/TaskListModel.scala
```diff
@@ -24,9 +24,10 @@ case class TaskListModel(taxYearSelection: Option[AccountingYearModel],
                          selfEmployments: Seq[SelfEmploymentData],
                          selfEmploymentAccountingMethod: Option[AccountingMethod],
                          ukProperty:  Option[PropertyModel],
-                         overseasProperty: Option[OverseasPropertyModel]) {
+                         overseasProperty: Option[OverseasPropertyModel],
+                         permitTaxYearChange: Boolean = true) {
 
-  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(_.confirmed)
+  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(_.confirmed) || (taxYearSelection.nonEmpty && !permitTaxYearChange)
```
sb-dev, line 30: `taxYearSelection` is going to be empty so the `taxYearSelection.nonEmpty` check is not necessary.

aug24: I don't understand.  If `taxYearSelection` is empty then that would never be true.
What this does is consider the result "confirmed" which is usually a user action, if the tax year is actually being mandated.

app/views/agent/AgentTaskList.scala.html
aug24: Why is it?  It should return true if it has a value and the user is not permitted to change it.  That's precisely what the other comment is about.

aug24:
I realised that the editable flag could be moved into the tax year model.  That cleaned it up quite a bit.

app/models/common/TaskListModel.scala
```diff
@@ -18,40 +18,37 @@ package models.common
...
 
 case class TaskListModel(taxYearSelection: Option[AccountingYearModel],
                          selfEmployments: Seq[SelfEmploymentData],
                          selfEmploymentAccountingMethod: Option[AccountingMethod],
-                         ukProperty:  Option[PropertyModel],
+                         ukProperty: Option[PropertyModel],
                          overseasProperty: Option[OverseasPropertyModel]) {
 
-  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(_.confirmed)
+  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(a => !a.editable || a.confirmed)
```
sb-dev, line 28:
```suggestion
  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(taxYear => !taxYear.editable || taxYear.confirmed)
```

app/models/common/TaskListModel.scala
```diff
@@ -18,40 +18,37 @@ package models.common
...
 
-  val taxYearSelectedNotConfirmed: Boolean = taxYearSelection.exists(!_.confirmed)
+  val taxYearSelectedNotConfirmed: Boolean = taxYearSelection.exists(a => a.editable && !a.confirmed)
```
sb-dev, line 30:
```suggestion
  val taxYearSelectedNotConfirmed: Boolean = taxYearSelection.exists(taxYear => taxYear.editable && !taxYear.confirmed)
```

app/models/common/TaskListModel.scala
```diff
@@ -18,40 +18,37 @@ package models.common
...
                          overseasProperty: Option[OverseasPropertyModel]) {
 
-  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(_.confirmed)
+  val taxYearSelectedAndConfirmed: Boolean = taxYearSelection.exists(taxYear => !taxYear.editable || taxYear.confirmed)
 
-  val taxYearSelectedNotConfirmed: Boolean = taxYearSelection.exists(!_.confirmed)
+  val taxYearSelectedNotConfirmed: Boolean = taxYearSelection.exists(taxYear => taxYear.editable && !taxYear.confirmed)
 
-  val ukPropertyComplete: Boolean =  ukProperty.exists(_.confirmed)
+  val permitTaxYearChange: Boolean = taxYearSelection.exists(a => a.editable)
```
sb-dev, line 32:
```suggestion
  val permitTaxYearChange: Boolean = taxYearSelection.exists(_.editable)
```

sb-dev (approved):

PR #1147, opened 2023-01-11
Fetch eligibility by year and check if edit will be permitted.

aug24:
Will be rebased after 2012,1998 are merged.

### DIFF
--- a/it/helpers/IntegrationTestModels.scala
+++ b/it/helpers/IntegrationTestModels.scala
@@ -24,36 +24,21 @@ import models.common._
 import models.common.business._
 import models.common.subscription.EnrolmentKey
 import models.usermatching.UserDetailsModel
-import uk.gov.hmrc.domain.Generator
-import utilities.AccountingPeriodUtil
 
 import java.time.LocalDate
 
 object IntegrationTestModels {
 
-  /*
-   * this function returns a random nino each time it is called, if you need a constant nino use TestConstants.testNino
-   */
-  def newNino: String = new Generator().nextNino.nino
-
   val testStartDate: DateModel = DateModel.dateConvert(LocalDate.now)
-  val testOneDayAgo: DateModel = DateModel.dateConvert(LocalDate.now.minusDays(1))
+  private val testOneDayAgo: DateModel = DateModel.dateConvert(LocalDate.now.minusDays(1))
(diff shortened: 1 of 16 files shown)


test/controllers/agent/matching/ConfirmClientControllerSpec.scala
```diff
@@ -38,7 +37,6 @@ import services.mocks._
 import uk.gov.hmrc.http.{HttpResponse, InternalServerException}
 import utilities.HttpResult.HttpConnectorError
 import utilities.UserMatchingSessionUtil
-import utilities.UserMatchingSessionUtil.{firstName, lastName}
```
sb-dev, line 41: You can check that `ITSASessionKeys.ELIGIBLE_NEXT_YEAR_ONLY` has been added to the session in `the client is ineligible only for the current year`.

it/helpers/IntegrationTestModels.scala
```diff
@@ -65,15 +50,13 @@ object IntegrationTestModels {
     )
   )
   val testAccountingMethod: AccountingMethodModel = AccountingMethodModel(Cash)
-  val testAccountingMethodProperty: AccountingMethodPropertyModel = AccountingMethodPropertyModel(Cash)
-  val testAccountingMethodForeignProperty: OverseasAccountingMethodPropertyModel = OverseasAccountingMethodPropertyModel(Cash)
+  private val testAccountingMethodProperty: AccountingMethodPropertyModel = AccountingMethodPropertyModel(Cash)
   val testValidStartDate: DateModel = DateModel.dateConvert(LocalDate.now.minusYears(1))
   val testValidStartDate2: DateModel = DateModel.dateConvert(LocalDate.now.minusYears(2))
+  //noinspection ScalaStyle
```
sb-dev, line 56: Why?

aug24: IIRC correctly, because there were lots of magic numbers and it's only a test.

test/controllers/agent/matching/ConfirmClientControllerSpec.scala
aug24: Good point, done and pushed.

sb-dev (approved):

PR #1170, opened 2023-03-07
[ITSASU-2178] - Mandated next year what you need to do page

srikanthsunkara79:


app/views/individual/WhatYouNeedToDo.scala.html
```diff
@@ -42,9 +42,11 @@ <h1 class="govuk-heading-xl">@messages("what-you-need-to-do.heading")</h1>
   @if(onlyNextYear) {
     @nonMandatedNextYear
   } else if (mandatedCurrentYear){
-  @currentYearMandated
+    @currentYearMandated
+  } else if (mandatedNextYear){
```
AlexRimmerHMRC, line 46: This is not quite right, only should happen when mandated and eligible for next year only.

AlexRimmerHMRC (approved):

PR #1313, opened 2024-01-23
[ITSASU-2787-Update] - Confirmation page update

srikanthsunkara79:


conf/messages
```diff
@@ -383,7 +383,8 @@ sign-up-confirmation.final-declaration.heading                       = Submit a
 sign-up-confirmation.final-declaration.para1                         = Other income sources should be declared, such as income from employment, dividends or savings. You need to report this income using either your:
 sign-up-confirmation.final-declaration.bullet1                       = compatible software (if it has the functionality)
 sign-up-confirmation.final-declaration.bullet2                       = HMRC online services account
-sign-up-confirmation.final-declaration.online-services.link-text     = You can find more information in your HMRC online services account.
+sign-up-confirmation.final-declaration.online-services.link-text     = You can find more information in your {0}.
+sign-up-confirmation.final-declaration.online-services.link          = HMRC online services account(opens in new tab)
```
AlexRimmerHMRC, line 387: we should not be manually defining opens in a new tab
remove the (opens in a new tab) from here, and instead use the helper for new tab links we have available in the twirl template, check other opens in new tab links for reference to how to use.

PR #1318, opened 2024-02-01
ITSASU-2808

sarinajsal:


app/views/agent/tasklist/addbusiness/YourIncomeSourceToSignUp.scala.html
```diff
@@ -52,13 +52,18 @@
         @messages("agent.your-income-source.caption", clientDetails.name, clientDetails.formattedNino)
     </span>
 
-    <h1 class="govuk-heading-xl">
-        @heading
-    </h1>
+    @if(incomeSources.hasNoIncomeSources){
```
AlexRimmerHMRC, line 55: I think this if statement needs to exist both here, but in the @heading function above, the heading function is used in multiple places and also determined the title of the page

Doing it this way will cause the heading and title of the page to be misaligned.

app/views/agent/tasklist/addbusiness/YourIncomeSourceToSignUp.scala.html
```diff
@@ -52,13 +58,18 @@
...
+        <h1 class="govuk-heading-xl">@heading</h1>
```
AlexRimmerHMRC, line 62: the h1's can now be moved above of this if statement, the condition is already handled by the heading functions if statement

```
@heading()
@if(...) {
 <p> blah blah </p>
} else {
 <p> other blah blah </p>
 }
```

AlexRimmerHMRC (approved):

PR #1344, opened 2024-03-28
[ITSASU-2973] - Control list kickout content change Individual

srikanthsunkara79:


conf/messages
```diff
@@ -120,13 +120,12 @@ cannot-use-service.title                                        = You cannot use
...
-cannot-take-part.title                                          = You cannot take part yet
-cannot-take-part.heading                                        = You cannot take part yet
-cannot-take-part.para1                                          = Making Tax Digital for Income Tax is not currently available people with certain types of income or who have not been trading long enough. You cannot take part yet.
-cannot-take-part.para2                                          = This service may be available to you in the future. Learn more about {0}.
-cannot-take-part.link1                                          = what makes you eligible to sign up
-cannot-take-part.inset1                                         = You must submit your Self Assessment tax return as normal.
+## Cannot sign up yet page
+cannot-sign-up-yet.title                                          = You cannot sign up yet
+cannot-sign-up-yet.heading                                        = You cannot sign up yet
+cannot-sign-up-yet.para1                                          = People with some types of income or deductions cannot sign up to Making Tax Digital for Income Tax (opens in new tab)
```
AlexRimmerHMRC, line 126: We should not be mentioning opens in new tab in our messages file, we have a component we can use NewTabLinkHelper or something like that which you should use instead. This helper should be used for all links which open in a new tab as it provides security features and also provides us with the additional text (eng + welsh) automatically.

app/views/individual/matching/NotEligibleForIncomeTax.scala.html
```diff
@@ -27,20 +27,17 @@
...
 
-@mainTemplate(title = messages("cannot-take-part.title"), backLink = Some("javascript:history.back()")) {
+@mainTemplate(title = messages("cannot-sign-up-yet.title")) {
 
-    <h1 class="govuk-heading-l">@messages("cannot-take-part.heading")</h1>
+    <h1 class="govuk-heading-l">@messages("cannot-sign-up-yet.heading")</h1>
 
-    <p class="govuk-body">@messages("cannot-take-part.para1")</p>
+   <p class="govuk-body">
+       <a class="govuk-link" href="https://www.gov.uk/guidance/check-if-youre-eligible-for-making-tax-digital-for-income-tax">@messages("cannot-sign-up-yet.para1")</a>,
```
AlexRimmerHMRC, line 35: See comment in messages file

ymabdulahi (approved):

PR #1426, opened 2024-10-11
[ITSASU-3352] - Create new page for Using Software - Individual

srikanthsunkara79:


app/controllers/individual/UsingSoftwareController.scala
```diff
@@ -0,0 +1,95 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.individual
+
+import auth.individual.SignUpController
+import config.AppConfig
+import forms.agent.UsingSoftwareForm
```
AlexRimmerHMRC, line 21:
```suggestion
import forms.individual.UsingSoftwareForm
```
Wrong form, always check imports are correct when copy + paste :D

PR #1433, opened 2024-10-21
ITSASU-3380:  Pre-pop - changes the card action link text when it's p…

tbg2003:
…re-popped - Agent


ILee15 (dismissed):
When changing the accounting method under Sole Trade Businesses but not confirming, the action link still says **Change** where I expected it to now be **Check**. It works correctly for the other sole trader details.

ILee15 (approved):

PR #1481, opened 2025-01-24
ITSASU-3586 - Updated agent uk property CYA to handle display of answ…

AlexRimmerHMRC:
…er when start date is before limit

### DIFF
--- a/app/views/agent/tasklist/ukproperty/PropertyCheckYourAnswers.scala.html
+++ b/app/views/agent/tasklist/ukproperty/PropertyCheckYourAnswers.scala.html
@@ -17,7 +17,9 @@
 @import models.common.PropertyModel
 @import views.html.templates.AgentMainTemplate
 @import utilities.UserMatchingSessionUtil.ClientDetails
+@import utilities.AccountingPeriodUtil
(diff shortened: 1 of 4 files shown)


app/views/agent/tasklist/ukproperty/PropertyCheckYourAnswers.scala.html
```diff
@@ -62,14 +64,28 @@
...
+                    }
+                )),
                 actions = Some(Actions(
                     items = Seq(
-                        ActionItem(
-                            href = controllers.agent.tasklist.ukproperty.routes.PropertyIncomeSourcesController.show(editMode = true, isGlobalEdit = isGlobalEdit).url,
-                            content = Text(messages(if(viewModel.startDate.isDefined) "base.change" else "base.add")),
-                            visuallyHiddenText = Some(messages("agent.property.check-your-answers.start-date.question"))
-                        )
+                        viewModel.startDateBeforeLimit match {
```
ILee15, line 75: The action items if they are the same href and visually hidden text could it just be something like
`content = Text(messages(if(viewModel.startDate.isDefined || viewModel.startDateBeforeLimit == Some(true)) "base.change" else "base.add")),`

ILee15 (approved):

PR #1515, opened 2025-03-13
ITSASU-3616: Created the individual income sources incomplete page

ymabdulahi:


it/test/helpers/ComponentSpecBase.scala
```diff
@@ -502,6 +502,10 @@ trait ComponentSpecBase extends AnyWordSpecLike with Matchers with OptionValues
         )
       )
     }
+    def showIncomeSourcesIncomplete(includeState: Boolean = true): WSResponse = get("/details/income-sources-incomplete", includeState = includeState)
```
ILee15, line 505: Do we want to start moving these methods into the ISpec directly?

ILee15 (approved):

PR #1528, opened 2025-03-26
ITSASU-3776: Back links for ORM page when EmailCaptureConsent feature…

ymabdulahi:
… switch enabled

### DIFF
--- a/app/controllers/agent/WhatYouNeedToDoController.scala
+++ b/app/controllers/agent/WhatYouNeedToDoController.scala
@@ -17,12 +17,15 @@
 package controllers.agent
 
 import config.AppConfig
+import config.featureswitch.FeatureSwitch.EmailCaptureConsent
+import config.featureswitch.FeatureSwitching
 import controllers.SignUpBaseController
 import controllers.agent.actions.{ConfirmedClientJourneyRefiner, IdentifierAction}
 import models.status.MandationStatus.Mandated
-import models.{Current, No, Yes}
+import models._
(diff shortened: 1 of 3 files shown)


app/controllers/agent/WhatYouNeedToDoController.scala
```diff
@@ -78,11 +86,25 @@ class WhatYouNeedToDoController @Inject()(view: WhatYouNeedToDo,
...
   }
 
-  def backUrl(eligibleNextYearOnly: Boolean, mandatedCurrentYear: Boolean): String = {
-    if (eligibleNextYearOnly || mandatedCurrentYear) {
-      controllers.agent.routes.UsingSoftwareController.show.url
+  def backUrl(eligibleNextYearOnly: Boolean, mandatedCurrentYear: Boolean, captureConsentStatus: Option[YesNo]): String = {
+    if (isEnabled(EmailCaptureConsent)) {
+      if (eligibleNextYearOnly) {
+        controllers.agent.routes.UsingSoftwareController.show.url
+      } else if (mandatedCurrentYear) {
```
ILee15, line 93: I am not sure this is correct, because it only accounts for users mandated for current year. 

So voluntary users signing up for current year get redirected to What Tax Year To Sign Up page. I think we only need to check if the tax year selection is Current as even if they are mandated the tax year is set to Current in `fetchSelectedTaxYear `

ILee15 (approved):

PR #1537, opened 2025-06-04
ITSASU-3950 add CST contact section to confirmation page

ILee15:


app/views/agent/confirmation/SignUpConfirmation.scala.html
```diff
@@ -67,6 +69,10 @@ <h2 class="govuk-heading-m">@messages("agent.sign-up-confirmation.what-must-do.h
 
     @agentServicesAccount
 
+    @if(usingSoftwareStatus && !taxYearSelectionIsNext) {
```
AlexRimmerHMRC, line 72: I'm not sure we need this usingSoftwareStatus condition, from what I remember, this flag was whether they selected yes or no to the software question, but since the change to kick them out, this is always true.

AlexRimmerHMRC (approved):

PR #1538, opened 2025-06-04
[ITSASU-3995] Refactor ClientAlreadySubscribedController

paddynski-opencast:


app/controllers/agent/matching/ClientAlreadySubscribedController.scala
```diff
@@ -16,33 +16,30 @@
...
 @Singleton
-class ClientAlreadySubscribedController @Inject()(val auditingService: AuditingService,
-                                                  val authService: AuthService,
+class ClientAlreadySubscribedController @Inject()(identify: IdentifierAction,
                                                   clientAlreadySubscribed: ClientAlreadySubscribed)
-                                                 (implicit val ec: ExecutionContext,
-                                                  val appConfig: AppConfig,
-                                                  mcc: MessagesControllerComponents) extends UserMatchingController {
+                                                 (val config: Configuration, val env: Environment)
+                                                 (implicit mcc: MessagesControllerComponents) extends SignUpBaseController with AuthRedirects {
```
AlexRimmerHMRC, line 33:
```suggestion
                                                 (implicit mcc: MessagesControllerComponents) extends SignUpBaseController {
```
I don't think AuthRedirects is used anywhere in this controller, so should be fine to remove.
If it is removed, config and env can also go as they are only needed due to AuthRedirects

test/controllers/agent/matching/ClientAlreadySubscribedControllerSpec.scala
AlexRimmerHMRC, line 1: Given the simplicity of the controller updated, I would actually prefer this controller spec to be deleted.
We would want to add a test into the ISpec to check authentication in that case.

app/controllers/agent/matching/ClientAlreadySubscribedController.scala
```diff
@@ -16,33 +16,30 @@
...
 
-  val show: Action[AnyContent] = Authenticated.async { implicit request =>
-    _ =>
-      Future.successful(Ok(clientAlreadySubscribed(
-        postAction = controllers.agent.matching.routes.ClientAlreadySubscribedController.submit
-      )))
+  val show: Action[AnyContent] = identify.async { implicit request =>
```
AlexRimmerHMRC, line 35: Can be simplified to remove the `Future.successful(...)` wrapper of the result and just use `identify { implicit request =>` instead

Same for submit

test/controllers/agent/matching/ClientAlreadySubscribedControllerSpec.scala
paddynski-opencast: Removed spec and added test to ispec.

PR #1539, opened 2025-06-04
ITSASU-3996: Updated ClientDetailsErrorController with new action con…

ymabdulahi:
…trollers


app/controllers/agent/matching/ClientDetailsErrorController.scala
```diff
@@ -16,25 +16,22 @@
...
                                              clientDetailsError: ClientDetailsError)
                                             (implicit val ec: ExecutionContext,
-                                             val appConfig: AppConfig,
-                                             mcc: MessagesControllerComponents) extends UserMatchingController {
+                                             mcc: MessagesControllerComponents) extends SignUpBaseController {
 
-  def show: Action[AnyContent] = Authenticated.async { implicit request =>
-    _ => Future.successful(Ok(clientDetailsError()))
+  def show: Action[AnyContent] = identify.async { implicit request =>
+    Future.successful(Ok(clientDetailsError()))
```
ILee15, line 34: It doesn't really need to be `.async` so you could remove that alongside `Future.successful`

test/controllers/agent/matching/ClientDetailsErrorControllerSpec.scala
```diff
@@ -17,16 +17,16 @@
...
 import org.mockito.Mockito.when
 import play.api.http.Status
 import play.api.mvc.{Action, AnyContent}
 import play.api.test.Helpers.{contentType, _}
 import play.twirl.api.HtmlFormat
-import services.mocks.MockAuditingService
 import views.html.agent.matching.ClientDetailsError
 
-class ClientDetailsErrorControllerSpec extends AgentControllerBaseSpec with MockAuditingService {
+class ClientDetailsErrorControllerSpec extends AgentControllerBaseSpec with MockIdentifierAction {
```
ILee15, line 29: As the controller is so simple, we could remove these unit tests as we have the ISpec that tests the same thing. Could you also add unauthorised tests in the ISpec please?

app/controllers/agent/matching/ClientDetailsErrorController.scala
```diff
@@ -16,25 +16,22 @@
 
 package controllers.agent.matching
 
-import auth.agent.UserMatchingController
-import config.AppConfig
+import controllers.SignUpBaseController
+import controllers.agent.actions.{ConfirmedClientJourneyRefiner, IdentifierAction}
```
ILee15, line 20: unused import here :)

PR #1543, opened 2025-06-09
ITSASU-3733 hidden banner and buttons during print preview

Ashgeo86:


app/views/individual/GlobalCheckYourAnswers.scala.html
```diff
@@ -75,7 +75,8 @@ <h2 class="govuk-heading-m">@messages("individual.global-check-your-answers.befo
         @saveAndContinueButtonHelper(
             isIndividual = true,
             saveAndContinueText = Some(messages("base.confirm-and-continue")),
-            saveAndReturnReference = Some("global-check-your-answers")
+            saveAndReturnReference = Some("global-check-your-answers"),
+            additionalClasses = Some("govuk-!-display-none-print")
```
AlexRimmerHMRC, line 79: Could it be better, instead of providing this class in this one specific case, having it be added all the time?
The parameter would be removed and the classes in the helper would just always be applied.

PR #1552, opened 2025-08-08
Remove accounting method from CYA view

ian-ralphs:


test/models/common/subscription/CreateIncomeSourcesModelSpec.scala
```diff
@@ -35,7 +35,7 @@ class CreateIncomeSourcesModelSpec extends PlaySpec with MustMatchers {
 
   val fullSoleTraderBusinesses: SoleTraderBusinesses = SoleTraderBusinesses(
     accountingPeriod = AccountingPeriodModel(now, now),
-    accountingMethod = Cash,
+    accountingMethod = Some(Cash),
```
ILee15, line 38: I think it would be good to add tests for when `accountingMethod = None` in each income source in this model spec

ILee15 (approved):

PR #1565, opened 2025-09-24
Itsasu 4375

bkarimii:


app/forms/validation/utils/ConstraintUtil.scala
```diff
@@ -32,7 +34,33 @@ object ConstraintUtil {
           case r => r
         }
       )
+  }
 
+  def isAfter(
```
ILee15, line 39: I think these would be better put into the Constraints object `forms/validation/Constraints.scala`

ILee15 (commented):
There is a bit of code in `DateHelper.scala.html` which was kept to support the use of the old dateModelMapping you've just replaced. Should be able to remove the if statement from here without any side effects:

```
            if(dateFields.exists(field => errorKey.contains(field))) { //DateModelMapping uses error message keys to highlight fields.
                if(errorKey.contains(key)) "govuk-input--error" else ""
            } else {
                error.map(_.args) match {
                    case Some(Seq()) => "govuk-input--error"
                    case Some(args) if args.contains(key) => s"govuk-input--error"
                    case _ => ""
                }
            }
```

bkarimii:
> There is a bit of code in `DateHelper.scala.html` which was kept to support the use of the old dateModelMapping you've just replaced. Should be able to remove the if statement from here without any side effects:
> 
> ```
>             if(dateFields.exists(field => errorKey.contains(field))) { //DateModelMapping uses error message keys to highlight fields.
>                 if(errorKey.contains(key)) "govuk-input--error" else ""
>             } else {
>                 error.map(_.args) match {
>                     case Some(Seq()) => "govuk-input--error"
>                     case Some(args) if args.contains(key) => s"govuk-input--error"
>                     case _ => ""
>                 }
>             }
> ```

I tried to remove it, but once you delete the `if-else` statement, the method becomes empty, so you have to delete that as well. And in the end, you also need to delete @InputClass. It's like a sequence , and deleting all of them will change style of the date fields in the browser. or I might be approaching this the wrong way.

PR #1701, opened 2026-04-08
[ITSASU-4949]

paddynski-opencast:


app/config/frontendAppConfig.scala
```diff
@@ -121,6 +121,8 @@ trait AppConfig {
 
   def getVAndCUrl: String
 
+  def getSessionHost: String
```
AlexRimmerHMRC:
```suggestion
  def getIncomeTaxSessionDataHost: String
```
Not saying it must be this, but as it seems a bit vague.

ymabdulahi (approved):

PR #1714, opened 2026-04-22
ITSASU-5015: Created the loading spinning page for confirming sign up

ymabdulahi:


app/views/individual/LoadingSpinner.scala.html
```diff
@@ -0,0 +1,85 @@
...
+mainTemplate: PrincipalMainTemplate,
+form: FormWithCSRF
+)
+
+@()(implicit request: Request[_], messages: Messages)
+
+@mainTemplate(title = messages("loading-spinner.heading")) {
+
+    <noscript>
+        <meta http-equiv="refresh" content="10" />
```
paddynski-opencast: This will not work if JS is enabled as there is no JS equivalent.
Please remove from `<noscript>` section.

app/views/agent/ClientLoadingSpinner.scala.html
```diff
@@ -0,0 +1,83 @@
...
+        mainTemplate: AgentMainTemplate,
+        form: FormWithCSRF
+)
+
+@()(implicit request: Request[_], messages: Messages)
+
+@mainTemplate(title = messages("loading-spinner.heading")) {
+
+    <noscript>
+        <meta http-equiv="refresh" content="10" />
```
paddynski-opencast, line 30: This will not work if JS is enabled as there is no JS equivalent.
Please remove from `<noscript>` section.

app/views/individual/LoadingSpinner.scala.html
AlexRimmerHMRC: I think it will be fine to leave in, it's not going live until it's fully built, which will include the integration and the javascript ideally.

paddynski-opencast (dismissed):

AlexRimmerHMRC (approved):

PR #1725, opened 2026-05-01
ITSASU-5064, updated hierarchies for incomeScourses page

WozCapGem:


paddynski-opencast (commented):
Please fix indentation in Twirl files.

app/views/agent/confirmation/SignUpConfirmation.scala.html
```diff
@@ -56,7 +56,7 @@
         @implicitDateFormatter.LongDate(signedUpDate).toLongDate
     </p>
 
-    <h2 class="govuk-heading-m">@messages("agent.sign-up-confirmation.what-must-do.heading")</h2>
+    <h2 class="govuk-heading-l">@messages("agent.sign-up-confirmation.what-must-do.heading")</h2>
```
AlexRimmerHMRC: I think this might be the wrong way round, by setting this to l, we are causing the same issue with the h1 and this h2.

I think instead, we need to take a look at the page headings and ensure that all h2s are smaller than the h1, all h3s are smaller than the h2s and so on.

AlexRimmerHMRC (approved):

PR #1744, opened 2026-05-21
ITSASU-5143

NoorMhd7:
To align with the approach taken in the `IVSuccessController` migration, the ISpec has been expanded to include both flag scenarios. As the ISpec now covers all scenarios previously tested in the unit spec, the unit spec has been deleted. `ivFailure` in `IncomeTaxSubscriptionFrontend` has also been updated to accept `sessionKeys` to support the new test cases.


app/controllers/individual/iv/IVFailureController.scala
```diff
@@ -31,28 +32,28 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class IVFailureController @Inject()(val authService: AuthService,
```
paddynski-opencast: Plrsde remove `authService` as it is n not required

app/controllers/individual/iv/IVFailureController.scala
```diff
@@ -31,28 +32,28 @@ import scala.concurrent.{ExecutionContext, Future}
...
                                     val auditingService: AuditingService,
```
paddynski-opencast: You can remove the `val` here.

app/controllers/individual/iv/IVFailureController.scala
```diff
@@ -31,28 +32,28 @@ import scala.concurrent.{ExecutionContext, Future}
...
+                                    identify: IdentifierAction,
                                     ivFailure: IVFailure)
                                    (implicit val ec: ExecutionContext,
                                     val appConfig: AppConfig,
```
paddynski-opencast: No need for this, please remove.

app/controllers/individual/iv/IVFailureController.scala
```diff
@@ -31,28 +32,28 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class IVFailureController @Inject()(val authService: AuthService,
                                     val auditingService: AuditingService,
+                                    identify: IdentifierAction,
```
paddynski-opencast: You cannot use `IdentifierAction` because it checks that user has the right confidence level abd this controller is for the case where the user has the wrong confidence level.

You have to use `BasicIdentifierAction` instead.

paddynski-opencast (approved):

PR #1749, opened 2026-05-22
ITSASU-5148

NoorMhd7:


app/controllers/individual/matching/CannotUseServiceController.scala
```diff
@@ -16,28 +16,23 @@
...
 import javax.inject.{Inject, Singleton}
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
-class CannotUseServiceController @Inject()(val auditingService: AuditingService,
-                                           val authService: AuthService,
-                                           val cannotUseServiceView: CannotUseService)
+class CannotUseServiceController @Inject()(cannotUseServiceView: CannotUseService)
                                           (implicit val ec: ExecutionContext,
                                            val appConfig: AppConfig,
```
paddynski-opencast: No need for `sppConfig` so please remove.

app/controllers/individual/matching/CannotUseServiceController.scala
```diff
@@ -16,28 +16,23 @@
...
-                                           mcc: MessagesControllerComponents) extends StatelessController {
+                                           mcc: MessagesControllerComponents) extends SignUpBaseController {
 
-  val show: Action[AnyContent] = Authenticated.asyncUnrestricted { implicit request =>
-    _ =>
-      Future.successful(Ok(cannotUseServiceView(
+  val show: Action[AnyContent] = Action { implicit request =>
```
paddynski-opencast: I thiunk we need `identify` instead of `Action` as we need to make sure the user is lkogged-in. Please check with Alex.

If you do change this then please add a test for unauthorised to the `ISpec`.

paddynski-opencast (approved):

PR #1754, opened 2026-06-01
ITSASU-4990: Added service url to SPS Handoff pages for preferences f…

ymabdulahi:
…rontend (P1)

### DIFF
--- a/app/controllers/individual/claimenrolment/sps/SPSHandoffForClaimEnrolController.scala
+++ b/app/controllers/individual/claimenrolment/sps/SPSHandoffForClaimEnrolController.scala
@@ -42,7 +42,8 @@ class SPSHandoffForClaimEnrolController @Inject()(
         _ =>
           goToSPS(returnUrl = appConfig.baseUrl + controllers.individual.claimenrolment.sps.routes.SPSCallbackForClaimEnrolController.callback,
             returnLinkText = "I have verified",
-            regime = "itsa"
+            regime = "itsa",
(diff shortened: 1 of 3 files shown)


app/controllers/individual/claimenrolment/sps/SPSHandoffForClaimEnrolController.scala
```diff
@@ -42,7 +42,8 @@ class SPSHandoffForClaimEnrolController @Inject()(
         _ =>
           goToSPS(returnUrl = appConfig.baseUrl + controllers.individual.claimenrolment.sps.routes.SPSCallbackForClaimEnrolController.callback,
             returnLinkText = "I have verified",
-            regime = "itsa"
+            regime = "itsa",
+            serviceUrl = appConfig.govukGuidanceITSASignUpAgentLink
```
paddynski-opencast: Why does that say `Agent`?
Isn't SPS for Individual only?

ymabdulahi: It was in the comment in the ticket for making the change for SPSHandoffForClaimEnrolController:
**redirectToSPS - pass serviceUrl = appConfig.govukGuidanceITSASignUpAgentLink to goToSPS**
I recognised it too, forgot to make that change, i've re-ran the test and checked it manually and it works as expected :)

paddynski-opencast (approved):

PR #1755, opened 2026-06-01
ITSASU-5167-f, bug fix

WozCapGem:


test/views/agent/eligibility/CannotTakePartViewSpec.scala
```diff
@@ -224,6 +224,11 @@ class CannotTakePartViewSpec extends ViewSpec {
       paragraph.text mustBe NoDataFound.paragraph1
     }
 
+    "have an inset paragraph with a link" in {
+      val paragraph: Element = document.mainContent.selectNth("#main-content > div > div > div.govuk-inset-text", 1)
+      paragraph.text mustBe NoDataFound.indentText
```
AlexRimmerHMRC, line 229: identText does not exist, so does not run.

If we are testing an inset paragraph with a link, we should probably also test the details of that link.

```
link.text mustBe ""
link.attr("href") mustBe ""
anything else you need to test, like if it opens in a new tab etc.
```

same for the other test you updated

AlexRimmerHMRC (approved):

PR #1775, opened 2026-06-12
[ITSASU-5211]

paddynski-opencast:


app/services/individual/UpsertAndAllocateEnrolmentService.scala
```diff
@@ -54,11 +55,17 @@ class UpsertAndAllocateEnrolmentService @Inject()(taxEnrolmentsConnector: TaxEnr
...
 
-  private def upsertEnrolment(enrolmentKey: EnrolmentKey, nino: String)(implicit hc: HeaderCarrier) = {
-    val enrolmentVerifiers = EnrolmentVerifiers(NINO -> nino)
-    taxEnrolmentsConnector.upsertEnrolment(enrolmentKey, enrolmentVerifiers) map {
-      case Right(value) => Right(value)
-      case Left(_) => Left(UpsertKnownFactsFailure)
+  def upsertEnrolment(enrolmentKey: EnrolmentKey, nino: String)(implicit hc: HeaderCarrier) = {
+    if (isEnabled(DistributedKnownFactsPattern)) {
+      Future.successful(Right(
+        UpsertEnrolmentResponseHttpParser
```
AlexRimmerHMRC: I think this line needs updating, you are returning the full parser as the right response. This doesn't break anything as we don't actually use the result of the right, but it's a type issue and I believe if you let your IDE generate the type for this function you would notice the problem more.
You probably meant to use the success response from that parser instead.

paddynski-opencast: Addressed.

AlexRimmerHMRC (approved):

PR #1793, opened 2026-07-13
[ITSASU-5221] Refactor individual.ProgressSavedController

paddynski-opencast:


app/controllers/individual/tasklist/ProgressSavedController.scala
```diff
@@ -39,37 +40,33 @@ class ProgressSavedController @Inject()(progressSavedView: ProgressSaved,
...
-                  _ <- auditingService.audit(saveAndComebackAuditData)
-                } yield {
-                  Ok(progressSavedView(cacheExpiryDateProvider.expiryDateOf(timestamp.dateTime), signInUrl))
-                }
-              })
-            case None => throw new InternalServerException("[ProgressSavedController][show] - The last updated timestamp cannot be retrieved")
-          }
-        }
+  def show(location: Option[String] = None): Action[AnyContent] = (identify andThen refine).async { implicit request =>
+    referenceRetrieval.getIndividualReference(request.sessionData) flatMap { reference =>
```
AlexRimmerHMRC: Should be able to remove this and replace any use of reference with request.reference

AlexRimmerHMRC: Also check usage of nino service and utr service which are injected in this controller, with the swap, they may become redundant as can usually be pulled from the request

paddynski-opencast: Removed those two as they are redundant.

AlexRimmerHMRC (approved):

